PR #69, opened 2023-08-02
StakeAccountInfo endpoint, cbor metadata, transaction store without utxo store

satran004:
- Added endpoint to return StakeAccountInfo in Account aggregate.
- Added cbor to metadata endpoints
- Enhanced existing block by block number endpoint to support find by block hash
- Added a default UtxoClient impl in transaction store to enable transaction store to run without utxo store


stores/metadata/src/main/java/com/bloxbean/cardano/yaci/store/metadata/processor/MetadataProcessor.java
```diff
@@ -41,22 +45,27 @@ public void handleAuxDataEvent(AuxDataEvent auxDataEvent) {
...
                     jsonNode.fieldNames().forEachRemaining(fieldName -> {
+                        String labelCbor = labelToCborMap.get(fieldName);
                         TxMetadataLabel txMetadataLabel = TxMetadataLabel.builder()
                                 .slot(eventMetadata.getSlot())
                                 .txHash(txAuxData.getTxHash())
                                 .blockNumber(eventMetadata.getBlock())
                                 .blockTime(eventMetadata.getBlockTime())
                                 .label(fieldName)
                                 .body((jsonNode.get(fieldName)).toString())
+                                .cbor(labelCbor)
```
matiwinnetou, line 68: ❤️

stores/metadata/src/main/resources/db/migration/postgresql/V0_600_1__init.sql
```diff
@@ -7,6 +7,7 @@ create table transaction_metadata
     tx_hash               varchar(255) not null,
     label                 varchar(255),
     body                  text,
+    cbor                  text,
```
matiwinnetou, line 10: For me this is not a problem but what would happen if a developer would already have a db, they would have to index from zero right?

Question is when is it sensible to start using actual migrations.

aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/service/AccountService.java
```diff
@@ -24,9 +24,12 @@
 public class AccountService {
     private final LocalStateQueryClient localStateQueryClient;
 
-    private AccountService(LocalClientProvider localClientProvider) {
-        this.localStateQueryClient = localClientProvider.getLocalStateQueryClient();
-        log.info("Account Controller initialized >>>");
+    public AccountService(LocalClientProvider localClientProvider) {
```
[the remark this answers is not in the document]
satran004: Thanks. Added @Nullable

stores/metadata/src/main/resources/db/migration/postgresql/V0_600_1__init.sql
satran004: 0.0.11 has many breaking changes and need a full sync from beginning. In the pre-alpha phase (0.0.x), occasional full syncs might be necessary to minimize migration scripts. 

As we transition to beta (0.x.x) or at least one client app in production :), separate db migration scripts will be created for each release.

Issue #106, opened 2023-10-19
Add transaction witness data


satran004:
- Add support to store transaction witness data (VKeyWitness) in transaction store.

PR #115, opened 2023-10-30
feat: #106 Initial impl for transaction witness support

satran004:
- Store transaction witnesses
- Add endpoint to return witnesses by tx hash


matiwinnetou:
❤️

Sotatek-HuyLe3a (approved):

PR #141, opened 2023-11-24
Asset UTxOs Endpoint

edridudi:


stores-api/staking-api/src/main/java/com/bloxbean/cardano/yaci/store/api/staking/StakingApiConfiguration.java
```diff
@@ -5,10 +5,7 @@
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
-@ConditionalOnProperty(name = {"store.staking.enabled", "store.staking.api-enabled"},
-        havingValue = "true",
-        matchIfMissing = true
-)
+@ConditionalOnProperty(name = {"store.staking.enabled"}, havingValue = "true", matchIfMissing = true)
```
satran004, line 8: @edridudi Thanks for PR.

If I understand correctly, you are attempting to use the service class from the API module, and that is why conditional properties are being added at the controller level.

However, the current approach is to maintain only a standard interface for the storage layer. The 'service' layer classes are specific to the controller and may change in the future. Otherwise, we would need to maintain two sets of interfaces and ensure they are backward compatible.

I believe the custom application should rely on the storage interface, which provides domain objects, and build its own service layer if necessary.

Alternatively, the custom application can depend on API endpoints.

But after reviewing your changes, I am now open to standardizing the service layer. However, we need to do the following:

1. Review all service layer methods.
2. Maintain the API level flag name at the store level. So, if api-enabled is false, both service and controllers are disabled. 
The api-enabled flag is at the store level, so it would always be ``store.<store_name>.api-enabled``.
3. Introduce another flag (or flags) for endpoints/controllers of the store API module, which would be at the controller level.

    There are two possible options:
    - **Option 1:** ``store.<store_name>.endpoints.<api_tag_name>.enabled = true``
    In this case, it's clear that the property will be defined in the specific store's starter.
    - **Option 2:** ``store.endpoint.<api_tag_name>.enabled = true``
    In this case, I am not sure what is the best starter to keep this property, as we have the same API tag in multiple stores.

    **Can we go with Option 1 for now because there is no ambiguity?** But we can explore Option 2 in future releases.

stores-api/utxo-api/src/main/java/com/bloxbean/cardano/yaci/store/api/utxo/controller/AssetController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final AssetService assetService;
+
+    @Autowired
+    public AssetController(AssetService assetService) {
+        this.assetService = assetService;
+    }
+
+    @GetMapping("/utxos/unit/{unit}")
+    @Operation(summary = "Asset UTxOs by Unit", description = "Get all UTxOs information of an asset.")
```
satran004, line 30: This looks good to me.

stores/utxo/src/main/java/com/bloxbean/cardano/yaci/store/utxo/storage/impl/UtxoStorageReaderImpl.java
```diff
@@ -54,6 +55,27 @@ public List<AddressUtxo> findUtxoByAddress(@NonNull String address, int page, in
...
+    public List<AddressUtxo> findUtxosByAsset(String unit, int page, int count, Order order) {
+        Pageable pageable = PageRequest.of(page, count)
+                .withSort(order.equals(Order.desc) ? Sort.Direction.DESC : Sort.Direction.ASC, "slot", "txHash", "outputIndex");
+
+        var query = dsl
+                .select(ADDRESS_UTXO.fields())
+                .from(ADDRESS_UTXO)
+                .leftJoin(TX_INPUT)
+                .using(field(ADDRESS_UTXO.TX_HASH), field(ADDRESS_UTXO.OUTPUT_INDEX))
+                .where(field(ADDRESS_UTXO.AMOUNTS).cast(String.class).contains(unit))
```
satran004, line 68: @edridudi Fyi, We are currently using a workaround by checking assets through the contains method, but this may create some edge case scenarios, such as when there are two assets with the same policy and a similar prefix in their names.

I understand that other queries in this class are following the same approach. However, I think we should plan to change this query to utilize JOOQ's JSON query support. We may want to look into that in this PR or we can plan it in a separate PR.

satran004 (changes requested):

stores/utxo/src/main/java/com/bloxbean/cardano/yaci/store/utxo/storage/impl/UtxoStorageReaderImpl.java
edridudi: changed this to equalIgnoreCase

stores/utxo/src/main/java/com/bloxbean/cardano/yaci/store/utxo/storage/impl/UtxoStorageReaderImpl.java
```diff
@@ -54,6 +55,27 @@ public List<AddressUtxo> findUtxoByAddress(@NonNull String address, int page, in
...
+    public List<AddressUtxo> findUtxosByAsset(String unit, int page, int count, Order order) {
+        Pageable pageable = PageRequest.of(page, count)
+                .withSort(order.equals(Order.desc) ? Sort.Direction.DESC : Sort.Direction.ASC, "slot", "txHash", "outputIndex");
+
+        var query = dsl
+                .select(ADDRESS_UTXO.fields())
+                .from(ADDRESS_UTXO)
+                .leftJoin(TX_INPUT)
+                .using(field(ADDRESS_UTXO.TX_HASH), field(ADDRESS_UTXO.OUTPUT_INDEX))
+                .where(field(ADDRESS_UTXO.AMOUNTS).cast(String.class).equalIgnoreCase(unit))
```
satran004, line 68: @edridudi This may not work as AMOUNTS is a json array. So exact matching will not work.
Let's keep it as "contains" for now.

We will create another issue to fix this. We need to use json functions from JOOQ https://www.jooq.org/doc/latest/manual/sql-building/column-expressions/json-functions/json-attribute-access/

satran004 (approved):

PR #149, opened 2023-11-29
UTxOs Storage Fixes

edridudi:


satran004 (approved):
Thanks @edridudi . Looks good me.

To determine the safest block number to delete from, you can use the 'store.cardano.cursor-no-of-blocks-to-keep' property. This property already exists with a default value of 2160, which is used for cursor cleanup. On the mainnet, changes become immutable after 2160 blocks. This is a configuration of the network.

PR #164, opened 2024-01-09
test: add tests for mir store processor

Sotatek-HuyLe3a:


stores/mir/src/test/resources/application.yml
```diff
@@ -0,0 +1,43 @@
+spring:
```
satran004, line 1: Thanks @Sotatek-HuyLe3a . Do we need the following commented properties? If not, we can remove.

satran004 (approved):

PR #195, opened 2024-02-07
Add epochNo,blockHeight,blockTime Fields to Utxo Object

edridudi:


components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/domain/Utxo.java
```diff
@@ -8,30 +8,34 @@
...
+
     private String txHash;
     private int outputIndex;
     private String address;
     private List<Amount> amount;
     private String dataHash;
     private String inlineDatum;
     private String referenceScriptHash;
+    private Integer epochNo;
+    private Long blockHeight;
```
satran004, line 31: Should we rename it to "blockNumber", as most of the domain classes use "blockNumber" ?
Similarly, we can use "epoch" instead of "epochNo".

satran004 (commented):
Thanks @edridudi .

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/domain/Utxo.java
edridudi: Sure

satran004 (approved):

Issue #204, opened 2024-02-28
(Preview) Block time, epoch slot are slightly off by a few seconds during syncing from genesis


Sotatek-HuyLe3a:
Reason: startSlot value is retrieved from the era table incorrectly

PR #205, opened 2024-02-28
Fix incorrect block time and epoch slotno error in preview

Sotatek-HuyLe3a:
#204


satran004 (approved):
Thanks @Sotatek-HuyLe3a

PR #217, opened 2024-03-13
chore: add address table

Sotatek-HuyLe3a:


aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/storage/impl/AccountBalanceStorageImpl.java
```diff
@@ -80,18 +84,25 @@ public List<AddressBalance> getAddressBalance(String address) {
     @Transactional
     @Override
     public void saveAddressBalances(@NonNull List<AddressBalance> addressBalances) {
-        List<AddressBalanceEntity> entities = addressBalances.stream().map(mapper::toAddressBalanceEntity)
+        List<AddressBalanceEntity> addressBalanceEntities = addressBalances.stream().map(mapper::toAddressBalanceEntity)
                 .toList();
+        List<AddressEntity> addressEntities = addressBalances.stream().map(mapper::toAddressEntity)
+                .collect(Collectors.toMap(AddressEntity::getAddress, Function.identity(), (a,b) -> a))
+                .values()
```
satran004, line 91: Or, should we just collect it in a Set for distinct values?

aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/storage/impl/AccountBalanceStorageImpl.java
```diff
@@ -153,6 +160,37 @@ private void saveAddrBalanceBatchJOOQ(List<AddressBalanceEntity> addressBalanceE
 
     }
 
+    private void saveAddressBatch(List<AddressEntity> addressEntities) {
+        if (enableJPAInsert) {
+            //TODO
```
satran004, line 165: I think we can ignore enableJPA for address. Let's just JOOQ

aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/storage/impl/AccountBalanceStorageImpl.java
```diff
@@ -153,6 +160,37 @@ private void saveAddrBalanceBatchJOOQ(List<AddressBalanceEntity> addressBalanceE
...
+        } else {
+            if (log.isTraceEnabled())
+                log.trace("Inserting address using JOOQ batch");
+
+            transactionTemplate.execute(status -> {
```
satran004, line 170: Probably we don't need a new transaction for address saving as # of addresses should not be too much.

aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/storage/impl/AccountBalanceStorageImpl.java
satran004: Can we just pass List of AddressBalance to saveAddresses method and handle it there?
As we will be using only JOOQ for address saving, we don't need to convert it to AddressEntity.

Something like this I had done before to track address
https://github.com/bloxbean/yaci-store/blob/41c4383a1db6c9bc4b87f21391c029e06fcdf266/aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/storage/impl/AccountBalanceStorageImpl.java#L159

satran004: Fyi. Now there is a AddressUtil class in account module which can give address and full address for an address. So we can use it to populate full address.

satran004 (approved):

Issue #222, opened 2024-03-18
Missing pre-defined DReps (ABSTAIN, NO_CONFIDENCE) in "delegation_vote" table


satran004:
DelegationVoteProcessor is currently storing  DRep hash and DrepId. But it doesn't store pre-defined dreps ABSTAIN, NO_CONFIDENCE.

**Solution:**

- Add a new column drep_type 
- Store all possible types (KEYHASH, SCRIPTHASH, ABSTAIN, NO_CONFIDENCE)

PR #226, opened 2024-03-19
chore: #222 add a new column drep_type

Sotatek-HuyLe3a:


Sotatek-HuyLe3a:
Old code still store pre-defines DRep, so I just add new column for DrepType

satran004 (approved):

PR #251, opened 2024-04-19
chore: add new flag to save address_tx_amount records that have zero token amount

Sotatek-HuyLe3a:


aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/processor/AddressTxAmountProcessor.java
```diff
@@ -162,6 +162,7 @@ private List<AddressTxAmount> processTxAmount(String txHash, EventMetadata metad
         return (List<AddressTxAmount>) addressTxAmountMap.entrySet()
                 .stream()
                 .filter(entry -> (accountStoreProperties.isAddressTxAmountIncludeZeroAmount() && entry.getKey().getSecond().equals(LOVELACE))
+                        || (accountStoreProperties.isAddressTxAmountIncludeZeroTokenAmount() && !entry.getKey().getSecond().equals(LOVELACE))
```
satran004, line 165: @Sotatek-HuyLe3a Do we need the condition !(LOVELACE) when IncludeZeroTokenAmount is set?

I think if IncludeZeroTokenAmount is set then we can include both lovelace and tokens, as token zero amount doesn't make sense without lovelace zero amount

What do you think ?

Sotatek-HuyLe3: Yes, doesn't make sense, but do you think if we set isAddressTxAmountIncludeZeroTokenAmount = true and isAddressTxAmountIncludeZeroAmount = false it will make  user confused?

satran004: Yes. I think it may create confusion.
So, we need to document it properly and set the default value.

Sotatek-HuyLe3a: Thank Satya, I think the second option is better option

satran004 (approved):

Issue #249, opened 2024-04-18
HealthCheckService:  Add last received block time


satran004:
Currently, the Healthcheck service only checks the connection status and error flag. However, we also need to verify whether the sync process is ongoing. This can be accomplished by checking the last received block time or the keepalive time.

PR #256, opened 2024-04-22
chore: (HealthService) Add last received block time

Sotatek-HuyLe3a:
#249

### DIFF
--- a/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/BlockFetchService.java
+++ b/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/BlockFetchService.java
@@ -21,7 +21,6 @@ import com.bloxbean.cardano.yaci.store.events.*;
 import io.micrometer.core.instrument.MeterRegistry;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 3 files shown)


components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/BlockFetchService.java
```diff
@@ -57,6 +59,10 @@ public class BlockFetchService implements BlockChainDataListener {
     private Integer previousEpoch;
     private Era previousEra;
 
+    @Setter
```
satran004, line 62: @Sotatek-HuyLe3a Do we need setter also? or only getter is enough to avoid any change from outside?

Sotatek-HuyLe3a: You're right, we need to remove Setter

Sotatek-HuyLe3a: I updated @satran004

satran004 (approved):
Thanks @Sotatek-HuyLe3a

Issue #255, opened 2024-04-21
[transaction store] Add a scheduled job to delete old transaction records


satran004:
- Introduce a TransactionPruningService (Same as UtxoPruningService)
- Delete transaction before a configured slot no

PR #258, opened 2024-04-23
feat: #225 add transaction pruning service

Sotatek-HuyLe3a:
#255

### DIFF
--- a/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
+++ b/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
@@ -1,0 +1,22 @@
+package com.bloxbean.cardano.yaci.store.transaction;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@Builder
+public class TransactionStoreProperties {
+
+    @Builder.Default
+    private boolean pruningEnabled = false;
+
+    @Builder.Default
+    private int pruningInterval = 144000;
+
+    @Builder.Default
+    private int pruningSafeSlot = 43200; // 2160 blocks
+}
--- a/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
+++ b/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
@@ -15,7 +15,7 @@ public class TransactionStoreProperties {
     private boolean pruningEnabled = false;
 
     @Builder.Default
-    private int pruningInterval = 144000;
+    private int pruningInterval = 86400;
 
     @Builder.Default
     private int pruningSafeSlot = 43200; // 2160 blocks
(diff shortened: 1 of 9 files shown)


stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/scheduler/TransactionPruningService.java
```diff
@@ -0,0 +1,84 @@
...
+    private final TransactionStoreProperties transactionStoreProperties;
+
+    private final AtomicBoolean isPruning = new AtomicBoolean(false);
+
+    @PostConstruct
+    public void init() {
+        log.info("<< Transaction Pruning Service Enabled >>");
+    }
+
+    @Scheduled(fixedRateString = "${store.transaction.pruning-interval:86400}", timeUnit = TimeUnit.MINUTES)
```
matiwinnetou, line 41: My suggestion is to keep the default or around 40 hours.

Sotatek-HuyLe3a: So you mean we should set default is 40 hours, right?

satran004: @matiwinnetou The current default 2160 (43200 slots) blocks is taken from the protocol parameter "securityParam" as no rollback possible after this duration. We are using the same value for other stores like utxo.

But I think, the application will set their own safe  slots based on application requirement.

stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
```diff
@@ -0,0 +1,22 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+@Builder
+public class TransactionStoreProperties {
+
+    @Builder.Default
+    private boolean pruningEnabled = false;
+
+    @Builder.Default
+    private int pruningInterval = 144000;
```
satran004, line 18: @Sotatek-HuyLe3a Let's keep this interval to 24hrs. 86400 sec.

stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Builder.Default
+    private int pruningSafeSlot = 43200; // 2160 blocks
```
satran004, line 21: I think @matiwinnetou  was referring to this value. 
But let's keep it as it is, as we are using similar value in other stores and it's configurable.

stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/scheduler/TransactionPruningService.java
```diff
@@ -0,0 +1,84 @@
...
+    private final TransactionStoreProperties transactionStoreProperties;
+
+    private final AtomicBoolean isPruning = new AtomicBoolean(false);
+
+    @PostConstruct
+    public void init() {
+        log.info("<< Transaction Pruning Service Enabled >>");
+    }
+
+    @Scheduled(fixedRateString = "${store.transaction.pruning-interval:144000}", timeUnit = TimeUnit.MINUTES)
```
satran004, line 41: Please change the time unit to second and the default value accordingly.

satran004 (approved):

PR #259, opened 2024-04-23
fix: Fix jooq for utxo pruning with h2

Sotatek-QuanLeA:


satran004:
@Sotatek-QuanLeA Thanks.

I tested the change with preprod network and all three databases (H2/Postgres/MySQL) without any error.

Let's try on mainnet before merging this change.

satran004 (approved):

PR #266, opened 2024-04-28
chore: Added an or condition for H2 where there is no space

satran004:
To fix utxo by asset query for H2. 
- Added another OR condition without space between "unit:" and unit value.


Sotatek-HuyLe3a (approved):

satran004:
Thanks @Sotatek-HuyLe3a

Issue #297, opened 2024-06-27
[Conway] Add 'constitution' table


Sotatek-HuyLe3a:
- Add 'constitution' table in governance store to store constitution info

PR #302, opened 2024-06-30
feat: add 'constitution' table

Sotatek-HuyLe3a:
#297


stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/domain/Constitution.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@SuperBuilder
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class Constitution {
```
satran004, line 17: Should we also add "epoch" ? May be activeEpoch.

stores/governance/src/main/resources/db/store/h2/V0_1100_1__init.sql
```diff
@@ -163,4 +163,17 @@ CREATE TABLE committee_member
 );
 
 CREATE INDEX idx_committee_member_slot
-    ON drep_registration (slot);
\ No newline at end of file
+    ON drep_registration (slot);
+
+CREATE TABLE constitution
```
satran004, line 168: If we are adding an epoch/activeEpoch, that could be a primary key? There will be only one possible Constitution in an epoch

PR #315, opened 2024-07-15
Store gov state from local query (WIP)

Sotatek-HuyLe3a:


stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/LocalGovActionStateProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+@ConditionalOnBean(LocalGovStateService.class)
+public class LocalGovActionStateProcessor {
```
satran004, line 21: Should we introduce storage layer for LocalxxxRepository ? Otherwise, in case of another storage impl for goveranance, we still need a SQL DB only for this part.

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/LocalGovActionStateProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+    private final LocalGovActionProposalStatusRepository localGovActionProposalStatusRepository;
+    private final LocalConstitutionRepository localConstitutionRepository;
+    private final LocalCommitteeMemberRepository localCommitteeMemberRepository;
+    private final LocalCommitteeRepository localCommitteeRepository;
+    private final LocalTreasuryWithdrawalRepository localTreasuryWithdrawalRepository;
+
+    @EventListener
+    @Transactional
+    public void handleEpochChangeEvent(EpochChangeEvent epochChangeEvent) {
+        if (!epochChangeEvent.getEventMetadata().isSyncMode())
```
satran004, line 32: What happens when we reach sync mode = true during initial sync, but we are still in the middle of the epoch ?

Should we somehow trigger this fetch once the sync reaches tip or close to tip even if the next epoch is not there yet ?

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/LocalGovActionStateProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+        localTreasuryWithdrawalRepository.deleteBySlotGreaterThan(slot);
+
+        if (count > 0) {
+            log.info("Fetch gov state after rollback event...");
+            localGovStateService.fetchAndSetGovState();
+        }
+    }
+
+    @Scheduled(fixedRate = 2, timeUnit = TimeUnit.MINUTES)
+    public void scheduleFetchAndSetCurrentProtocolParams() {
```
satran004, line 56: Is it going to handle the scenario mentioned in my previous comment ?

Does scheduler check if sync mode = true ? Also, do we need this scheduler?

stores/governance/src/main/resources/db/store/postgresql/V0_1100_1__init.sql
```diff
@@ -177,4 +177,61 @@ CREATE TABLE constitution
...
+(
+    threshold          double precision,
+    epoch              int,
+    slot               bigint,
+    update_datetime    timestamp,
+    PRIMARY KEY (epoch)
+);
+
+DROP TABLE IF EXISTS local_treasury_withdrawal;
+CREATE TABLE local_treasury_withdrawal
```
satran004, line 206: We need this data in reward calculation. Should we make it like instant_reward table ? (something like this in adapot branch) https://github.com/bloxbean/yaci-store/blob/3fc5c61a62a1db8be115779f5b967b120b9e3da7/aggregates/adapot/src/main/resources/db/store/postgresql/V0_1300_1__init.sql#L34

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/LocalGovActionStateProcessor.java
Sotatek-HuyLe3a: We will introduce storage layer for LocalxxxRepository

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/LocalGovActionStateProcessor.java
Sotatek-HuyLe3a: The reason I use the scheduler is that I have encountered a situation where, at the first block of an epoch, I performed a query to the node, but the result returned was similar to the previous epoch. After a few more blocks, I queried again, and the result returned had new information. Therefore, I need the scheduler to ensure accuracy.

However, I will recheck this to confirm whether the scheduler is really necessary or if there is a more reasonable alternative.

satran004: Thanks @Sotatek-HuyLe3a 

Can you please verify it again after the recent slot # fix in era table? We had similar issue in local_epoch_param table. It was due to wrong epoch no. But now it's ok

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalGovStateService.java
```diff
@@ -0,0 +1,345 @@
...
+        Integer currentEpoch = eraService.getEpochNo(com.bloxbean.cardano.yaci.core.model.Era.Conway, tip.getPoint().getSlot());
+        Long slot = tip.getPoint().getSlot();
+        var latestCursor = cursorService.getCursor();
+
+        if (latestCursor.isEmpty() || tip.getBlock() > latestCursor.get().getBlock() + MAX_BLOCK_DIFFERENCE) {
+            log.info("The max processed block is not close to the tip, ignore handling gov state");
+            return;
+        }
+
+        List<LocalGovActionProposalStatus> expiredAndRatifiedGovActionsInPrevEpoch = localGovActionProposalStatusStorageReader
```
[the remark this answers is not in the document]
Sotatek-HuyLe3a: Thank @satran004 , got it.  I will make it consistently.

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalGovStateService.java
```diff
@@ -0,0 +1,345 @@
...
+        List<GovActionProposal> enactedHardForkInitProposals = govActionProposalStorage.findByGovActionIds(enactedGovActionIds)
+                .stream()
+                .filter(govActionProposalEntity -> govActionProposalEntity.getType() == GovActionType.HARD_FORK_INITIATION_ACTION)
+                .toList();
+
+        localHardForkInitiationStorage.saveAll(
+                enactedHardForkInitProposals.stream()
+                        .map(govActionProposal -> {
+                            JsonNode actionDetail = govActionProposal.getDetails();
+                            JsonNode protocolVersionNode = actionDetail.get("protocolVersion");
```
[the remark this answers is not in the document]
Sotatek-HuyLe3a: I think in this case we don't need check null.
maybe we can't submit proposal with govaction detail is null, and it can't be ratified if its detail is null

satran004 (approved):

PR #318, opened 2024-07-16
[governance] Update migration script

Sotatek-HuyLe3a:
#319 

- Fix index issue: there are two indexes define for same column 'slot' (drep_registration)
- Add "drop table if exists ..."
- Format SQL


satran004 (approved):

satran004:
Thanks @Sotatek-HuyLe3a. We can merge it for rc5 release.

PR #322, opened 2024-07-17
chore: Add other configuration page

satran004:


docs/pages/other-configurations.mdx
```diff
@@ -0,0 +1,66 @@
...
+```properties
+store.sync-auto-start=false
+```
+
+## Store specific application
+
+Yaci Store bundles all stores by default, and you can enable or disable stores by setting a store-specific enable flag in the application.properties file.
+This makes it possible to have multiple instances of yaci-store with different stores enabled. These individual instances can write to their own separate databases with the default configuration.
+
+However, if you want to have multiple store-specific yaci-store applications writing to the same database, each application needs to maintain its own **cursor** in the `_cursor` table.
```
Sotatek-HuyLe3a, line 60: "_cursor" ->  "cursor_"

Sotatek-HuyLe3a (approved):

PR #346, opened 2024-08-26
feat: handle local drep stake distribution and introduce new properties

Sotatek-HuyLe3a:
#342

### DIFF
--- a/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalDRepDistrService.java
+++ b/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalDRepDistrService.java
@@ -1,0 +1,148 @@
+package com.bloxbean.cardano.yaci.store.governance.service;
+
+import com.bloxbean.cardano.yaci.core.model.governance.Drep;
+import com.bloxbean.cardano.yaci.core.model.governance.DrepType;
+import com.bloxbean.cardano.yaci.core.protocol.chainsync.messages.Tip;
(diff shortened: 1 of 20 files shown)


stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalDRepDistrService.java
```diff
@@ -0,0 +1,148 @@
...
+                                    .drepHash(drepHash)
+                                    .drepType(drep.getType())
+                                    .amount(amount)
+                                    .epoch(epoch)
+                                    .slot(slot)
+                                    .build());
+                        });
+
+                        if (!localDRepDistrList.isEmpty()) {
+                            localDRepDistrStorage.saveAll(localDRepDistrList);
```
satran004, line 141: As this method is also called from scheduler, will it ignore saving the same data if it's already available for the epoch?

Sotatek-HuyLe3a: No, it will not ignore, in local_drep_distr, (drep_hash, epoch) is primary key, so if data' already available for the epoch, it will be updated

satran004: Should we ignore update if it's already there?

Sotatek-HuyLe3a: I removed scheduler, so it will not be updated

satran004 (approved):

PR #358, opened 2024-09-19
chore: update related to era value in fetching local data

Sotatek-HuyLe3a:


stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/service/LocalEpochParamService.java
```diff
@@ -44,9 +43,8 @@ public class LocalEpochParamService {
 
     private DomainMapper domainMapper = DomainMapper.INSTANCE;
 
-    @Value("${store.cardano.n2c-era:Conway}")
-    private String eraStr;
-
+    @Getter
+    @Setter
```
satran004, line 47: I think only getter is enough?

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/local/LocalGovActionStateProcessor.java
```diff
@@ -54,13 +55,19 @@ public void handleEpochChangeEvent(EpochChangeEvent epochChangeEvent) {
         if (!syncMode)
             return;
 
+        localGovStateService.setEra(Era.valueOf(epochChangeEvent.getEra().name()));
```
satran004, line 58: Can we move both blockEvent and epochEvent from this class to `LocalDRepDistrService` class to align with LocalEpochParam? Then we don't need to set era from outside. The processor can only have `rollback` method.

satran004: Also, let's try to avoid  a separate field for `syncMode` as this is already available in event metadata.

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalGovStateService.java
```diff
@@ -203,6 +186,11 @@ private void handleGovStateQueryResult(GovStateQueryResult govStateQueryResult)
                         .gov_action_index((int) entity.getGovActionIndex())
                         .build()).toList();
 
+        if (latestCursor.isEmpty() || tip.getBlock() > latestCursor.get().getBlock() + MAX_BLOCK_DIFFERENCE) {
```
satran004, line 189: Is this `if` condition required?

Sotatek-HuyLe3a: Because we need the content of the hard fork init and treasury withdrawal, we need to reference the gov_action_proposal table. 
However, we might not need this condition, since we only insert into the two tables local_hard_fork_initiation and local_treasury_withdrawal  when the actions are enacted, and the actions, from the time they are created to the time they are enacted, must take at least more than one epoch. Therefore, the condition syncMode = true might already be sufficient.

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/local/LocalGovActionStateProcessor.java
Sotatek-HuyLe3a: > Also, let's try to avoid a separate field for `syncMode` as this is already available in event metadata.
Yes, but in processor class, I think we still need 'syncMode' field since rollbackEvent don't have event metadata

satran004: Got it. Thanks.

stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/service/LocalGovStateService.java
satran004: If it's not required, let's remove it to keep the logic simple.

satran004 (approved):

Issue #365, opened 2024-09-26
Add Read-Only Mode Support to Yaci-Store


Sotatek-HuyLe3a:
To support true read-only mode, we propose to use the flag: store.read-only-mode.
When this flag is enabled, it will prevent any connection-related core beans from being created during startup (like BlockSync). This will allow the application to act independently as an API provider, without needing to connect to a Cardano node.

PR #366, opened 2024-09-26
feat: support read-only mode

Sotatek-HuyLe3a:
#365


satran004 (approved):
Thanks @Sotatek-HuyLe3a

Issue #373, opened 2024-10-07
[Script] Error log while trying to parse a native script in output's reference script field


satran004:
There is some test data in preprod where native scripts are attached to the output as reference scripts. The script store tries to parse the attached script reference bytes as a Plutus script, resulting in an error log when the parsing fails, which is expected.

A possible solution could be to detect the script type and parse it accordingly.

**Example:**

**Env:** Preprod
**Tx hash:** c3b701046b01330c7b28a680ca2d5a77a134d593fcccb489ae063b787b4587b0#1

```
c.b.c.y.store.script.helper.ScriptUtil   : Error deserializing script ref: 82008200581ca0dfc5656b946dea62a1fc23ff8881eb6468fdc14c295e2839c3ece3
2024-10-07T19:40:05.981+08:00 ERROR 11264 --- [Yaci Store App] [onPool-worker-4] c.b.c.y.store.script.helper.ScriptUtil   : Error deserializing plutus script in scriptRef : TxHash: c3b701046b01330c7b28a680ca2d5a77a134d593fcccb489ae063b787b4587b0#1, scriptRef: 82008200581ca0dfc5656b946dea62a1fc23ff8881eb6468fdc14c295e2839c3ece3
```

PR #378, opened 2024-10-11
fix: handle NativeScript as script ref

Sotatek-HuyLe3a:
#373 
- Updated code to handle native script case in ScriptRefProcessor
- Updated code to check for native scripts when finding Plutus Scripts (using getScripts in TxScriptFinder).
- Fixed test


stores/script/src/main/java/com/bloxbean/cardano/yaci/store/script/helper/TxScriptFinder.java
```diff
@@ -63,13 +65,15 @@ public Map<String, PlutusScript> getScripts(Transaction transaction) {
...
                         AddressUtxo addressUtxo = addressUtxoOptional.get();
                         if (addressUtxo.getScriptRef() != null && !addressUtxo.getScriptRef().isEmpty()) {
-                            PlutusScript plutusScript = ScriptUtil.deserializeScriptRef(addressUtxo);
-                            if (plutusScript != null)
-                                scriptsMap.put(ScriptUtil.getPlutusScriptHash(plutusScript), plutusScript);
-                            else {
-                                log.error("PlutusScript is null for tx: " + transaction.getTxHash());
-                            }
-                        }
+                            var cclScript = ScriptReferenceUtil.deserializeScriptRef(HexUtil.decodeHexString(addressUtxo.getScriptRef()));
```
satran004, line 68: @Sotatek-HuyLe3a What would happen if `deserializeScriptRef` throws an error? Ideally, we should just log the error and continue. Can you please verify?

Can we avoid double deserialization here? Since `ScriptReferenceUtil` is already deserializing to CCL's `PlutusScript`, perhaps we only need to handle the conversion to the store's `PlutusScript` inside a new conversion method in `ScriptUtil`

PR #398, opened 2024-11-13
feat: added get BlocksByEpochAPI + readded pages to blocks API

Kammerlo:
Hi,
I added a small API to the blocks controller to get Blocks by Epoch. We can discuss how the Endpoint should be called, currently just a placeholder.
Additionally I readded the Page to get Blocks. This allows users to easier query how many blocks are present at all. 

I assume you had a reason, why you removed the `Page` object and replaced it with `slice`. Was it due to performance?
I'm currently only letting this run on preprod, so can't say anything about speed on mainnet. 
I'll test soon.

### DIFF
--- a/stores-api/blocks-api/src/main/java/com/bloxbean/cardano/yaci/store/api/blocks/controller/BlockController.java
+++ b/stores-api/blocks-api/src/main/java/com/bloxbean/cardano/yaci/store/api/blocks/controller/BlockController.java
@@ -43,6 +43,16 @@ public class BlockController {
         }
     }
 
+    @GetMapping("epoch/{epoch}")
+    @Operation(summary = "Block List by Epoch", description = "Get blocks by epoch number.")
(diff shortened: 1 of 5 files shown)


satran004:
Thanks @Kammerlo 

We removed `Page` for performance reasons. Since `Page` performs an extra count query, it becomes very slow for networks like `mainnet`.

One of the goals is to keep the resource requirements minimal for default APIs, unless absolutely necessary.

Because block numbers are sequential, we can easily calculate the total number of blocks on the client side from the most recent block. Also, with `Slice`, only the previous and next pages are provided, which should be sufficient for the block list page. Typically, users search by block number or check the recent 20 blocks, then navigate to previous blocks as needed.

Hope that makes sense.

The new API looks good to me. The `Page` option might work well in this case due to the additional "epoch" filter, but we need to try it out. Also, we should decide if using both `Page` and `Slice` for different use cases is ok. It’s probably fine.

Kammerlo:
Reverted changes to slice like it was before + changed the new endpoint to slice as well. I'll calculate the page count in the frontend.

satran004 (approved):

PR #416, opened 2024-12-06
handle proposal refund with dbsync data & use governance rule in governance aggr

Sotatek-HuyLe3a:
- Handle proposal refund with dbsync data.
- Use  governance rule to check proposal status.
- Drop 'slot' column in governance_action_proposal_status.
- In the governance_action_proposal table, each epoch will store proposals with statuses of active, expired, or ratified.
- In drep_dist table, Use (drep_hash, epoch) as the primary key instead of (drep_id, epoch).
- Update drep dist snapshot query: 
    + With auto-abstain, no-confidence drep, drep_hash = 'ABSTAIN' or 'NO_CONFIDENCE' , instead of drep_id
    + Update the subquery for calculating active proposals.
- In governance-rule, update code to handle bootstrap phase or non bootstrap phase cases


satran004 (approved):
Thanks

Issue #431, opened 2025-02-03
Admin CLI - Manual Rollback command


satran004:
Add a new command to the `admin-cli` to roll back to the beginning of a previous epoch.

This command should perform the required validations and find the slot of the given epoch to update the cursor table.

The rollback should occur only to the last block of an epoch. This is necessary for accurate stake address balance calculation (when balance calculation is done in batches), which is used in reward calculation.

**For example:**
If rolling back to epoch x, the CLI should find the `slot` of the `last block` of epoch `x-1` and set the cursor accordingly in the `_cursor` table.

It should also delete records of all the existing tables based on `slot` no.

PR #436, opened 2025-02-12
Admin CLI - Manual Rollback command 

Sotatek-HuyLe3a:
#431


components/dbutils/src/main/resources/rollback.yml
```diff
@@ -0,0 +1,54 @@
+# asset store
```
satran004, line 1: Thanks @Sotatek-HuyLe3a 

I think we can simplify the yaml to something like this

```
tables:
  - assets
  - block
  - cost_model
  
```

Sotatek-HuyLe3a: @satran004, i updated  yaml file with simpler structure like above, you can check again

PR #456, opened 2025-02-25
Update Param Protocol after the PP change action's enacted

Sotatek-HuyLe3a:
- Update domain class GovActionProposal in common module and the related code
- Update Param Protocol after the PP change action's enacted

### DIFF
--- a/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/processor/EpochParamProcessor.java
+++ b/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/processor/EpochParamProcessor.java
@@ -1,10 +1,17 @@
 package com.bloxbean.cardano.yaci.store.epoch.processor;
 
 import com.bloxbean.cardano.yaci.core.model.Era;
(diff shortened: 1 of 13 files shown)


stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/processor/EpochParamProcessor.java
```diff
@@ -70,6 +78,22 @@ public void handleEpochChangeEvent(PreEpochTransitionEvent epochChangeEvent) {
         //Additional Era change specific rules
         ppEraChangeRules.apply(newEra, prevEra, protocolParams);
 
+        // handle parameter change from gov action proposal
+        if (epochChangeEvent.getEra() == Era.Conway) {
```
satran004, line 82: @Sotatek-HuyLe3a 

Instead of checking Conway era, we can check if the era is Conway era or later. You may need to get the value of the era and compare.

Sotatek-HuyLe3a: @satran004 , I just updated to support for later eras, please check again.

Issue #473, opened 2025-03-08
AaPot calculation status is set to COMPLETED even if the governance calculation or drep distribution query is not successful


satran004:
Currently, in the adapot background job, the job is marked as successful if the reward calculation is successful. But, we should also track the status of the governance calculation and the drep distribution snapshot status. If the drep distribution snapshot fails for any reason, we will get incorrect adapot calculation in future epochs.

Therefore, it’s crucial to mark the adapot job status as COMPLETE once all the steps of the epoch transitions are completed.

PR #477, opened 2025-03-13
 Fix adapot job status tracking and add drep distribution time (#473, #463)

Sotatek-HuyLe3a:
- Removed catching exception and reformat code in the ProposalStatusProcessor
- Added era checking before gov action status calc and drep dist calc
- Added drep distr query time to adapot_jobs table

### DIFF
--- a/aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/job/storage/AdaPotJobStorage.java
+++ b/aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/job/storage/AdaPotJobStorage.java
@@ -5,10 +5,13 @@ import com.bloxbean.cardano.yaci.store.adapot.job.domain.AdaPotJobStatus;
 import com.bloxbean.cardano.yaci.store.adapot.job.domain.AdaPotJobType;
 
 import java.util.List;
+import java.util.Optional;
 
 public interface AdaPotJobStorage {
 
     List<AdaPotJob> getJobsByTypeAndStatus(AdaPotJobType type, AdaPotJobStatus status);
+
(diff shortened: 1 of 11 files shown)


aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/job/storage/AdaPotJobStorage.java
```diff
@@ -5,11 +5,14 @@
 import com.bloxbean.cardano.yaci.store.adapot.job.domain.AdaPotJobType;
 
 import java.util.List;
+import java.util.Optional;
 
 public interface AdaPotJobStorage {
 
     List<AdaPotJob> getJobsByTypeAndStatus(AdaPotJobType type, AdaPotJobStatus status);
 
+    Optional<AdaPotJob> getJobByEpoch(int epoch);
```
satran004, line 14: @Sotatek-HuyLe3a Should we also include the type here? Currently, we have only one type of job reward_calc. May be we can use that for other types of job. Also, I think we may want to rename type REWARD_CALC to ADAPOT_CALC which covers everything. What do you think?

Sotatek-HuyLe3a: I think we should include the type. In the future, we may add other types.

Sotatek-HuyLe3a: @satran004 , i included the type, you can check again

satran004 (approved):

PR #483, opened 2025-03-19
[Governance] Implement DRep expiry handling

Sotatek-HuyLe3a:
#437

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
@@ -108,9 +108,10 @@ public class DRepExpiryProcessor {
                 .collect(Collectors.toMap(dRep -> new DRepInfo(dRep.getDrepHash(), dRep.getDrepId()), DRep::getSlot));
 
         // find dReps did vote in prev epoch
-        final Map<String, Long> dRepsVotedInPrevEpoch = votingAggrService.getVotesByVoterTypesInEpoch(List.of(VoterType.DREP_KEY_HASH, VoterType.DREP_SCRIPT_HASH), prevEpoch)
+        final List<String> dRepsVotedInPrevEpoch = votingAggrService.getVotesByVoterTypesInEpoch(List.of(VoterType.DREP_KEY_HASH, VoterType.DREP_SCRIPT_HASH), prevEpoch)
                 .stream()
-                .collect(Collectors.toMap(VotingProcedure::getVoterHash, VotingProcedure::getSlot));
+                .map(VotingProcedure::getVoterHash)
+                .collect(Collectors.toList());
 
         final List<DRepExpiry> dRepExpiryListInPrevSnapshot = dRepExpiryStorage.findByEpoch(prevEpoch - 1);
 
@@ -137,7 +138,7 @@ public class DRepExpiryProcessor {
             }
 
             // the case DRep is updated or DRep did vote in the previous epoch
-            if (updatedDRepsInPrevEpoch.get(dRepInfo) != null || dRepsVotedInPrevEpoch.get(dRepInfo.getDRepHash()) != null) {
+            if (updatedDRepsInPrevEpoch.get(dRepInfo) != null || dRepsVotedInPrevEpoch.contains(dRepInfo.getDRepHash())) {
                 newDRepExpiry.setActiveUntil(prevEpoch + drepActivity);
             }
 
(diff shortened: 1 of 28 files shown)


aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+
+        if (eraService.getEraForEpoch(event.getEpoch() - 1).getValue() < Era.Conway.getValue()) {
+            return;
+        }
+
+        int epoch = event.getEpoch();
+        int prevEpoch = epoch - 1;
+
+        // delete if existed
+        jdbcTemplate.update("delete from drep_expiry where epoch = :epoch", new MapSqlParameterSource().addValue("epoch", prevEpoch));
```
satran004, line 60: @Sotatek-HuyLe3a 
- Should we delete all records from drep_expiry or only create a record for a drep whose expiry changes? 
So, we need to get the latest record of the drep to get the correct expiry. I think we may need to keep a slot to handle the rollback scenario in that case. 
But, I agree that removing all records and creating a full snapshot for drep expiry for every epoch is probably a cleaner approach for now. We can optimize it later if it becomes slower.

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+                        .toList();
+
+        // The slot the first proposal was created slot in the previous epoch
+        Long firstProposalCreatedSlotInPrevEpoch = !govActionProposalsCreatedInPrevEpoch.isEmpty() ? govActionProposalsCreatedInPrevEpoch.get(0).getSlot() : null;
+
+        // get last epoch (before the prev epoch) that had ratified or active proposals
+        Optional<Integer> lastEpochsHadRatifiedOrActiveProposalsOpt = proposalStateClient.getLatestEpochWithStatusBefore(List.of(GovActionStatus.ACTIVE, GovActionStatus.RATIFIED), prevEpoch);
+
+        int dormantEpochsCount = lastEpochsHadRatifiedOrActiveProposalsOpt
+                .map(epochCount -> prevEpoch - epochCount - 1).orElseGet(() -> prevEpoch - 1);
```
satran004, line 92: Is the param name `epochCount` correct? Or it is `epoch`?
Similarly, prevEpoch - 1 in else condition is correct? It seems it represents a epoch no, not count.

aggregates/governance-aggr/src/main/resources/db/store/h2/V0_1400_1__init.sql
```diff
@@ -56,3 +56,17 @@ create table gov_action_proposal_status
     update_datetime    timestamp,
     primary key (gov_action_tx_hash, gov_action_index, epoch)
 );
+
+drop table if exists drep_expiry;
```
satran004, line 60: Can you add jooq entries for this new table in graalvm native image's resource-config.json file in this module? You can refer to other entries.. I think we need to add new generated pojo and table class.

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
Sotatek-HuyLe3a: > Is the param name epochCount correct? Or it is epoch?

Thanks, epochCount is not correct, it is epoch.

> Similarly, prevEpoch - 1 in else condition is correct? It seems it represents a epoch no, not count.

Yes, I handled this else case for scenarios where we use a devnet like devkit, where the initial epoch of the Conway era is 0. With this code implementation, the else case won’t occur on preprod, preview, or mainnet, because in the first epoch of the Conway era, there was already a governance action proposal. However, using prevEpoch - 1 is not general enough, we need to determine the first epoch of the Conway era.

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
Sotatek-HuyLe3a: >Should we delete all records from drep_expiry or only create a record for a drep whose expiry changes?
So, we need to get the latest record of the drep to get the correct expiry. I think we may need to keep a slot to handle the rollback scenario in that case.
But, I agree that removing all records and creating a full snapshot for drep expiry for every epoch is probably a cleaner approach for now. We can optimize it later if it becomes slower.

If we rely on the most recent snapshot and the DRep-related actions in the latest epoch, it benefits us in terms of rollback support, but it's less ideal in terms of data size,  the table will be slightly larger than the drep_dist table. Currently, the drep_dist table on mainnet is around 7MB in size.

However, I completely agree with you that when creating a snapshot — especially with DRep expiry data — it would be better if we only had a single snapshot, and this approach has the advantage of not depending on the previous snapshot. I’ll experiment with this.

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepExpiryProcessor.java
satran004: May be we can use `EraService` to get the conway era slot and then epoch.

satran004 (approved):

Issue #486, opened 2025-03-20
[governance] [preprod] DRep 'drep16pxnn38ykshfahwmkaqmke3kdqaksg4w935d7uztvh8y5l48pxv' should not included in DRep distr snapshot at the epoch boundary between epoch 175 and epoch 176 at the epoch boundary between epoch 175 and epoch 176


Sotatek-HuyLe3a:
In epoch 175, DRep 'drep16pxnn38ykshfahwmkaqmke3kdqaksg4w935d7uztvh8y5l48pxv' was deregistered and then re-registered. After that, it did not receive any further delegation votes.

Therefore, it should not appear in the DRep distribution snapshot at the epoch boundary between epoch 175 and epoch 176

PR #489, opened 2025-03-24
Add column 'registration_slot' to table 'drep_dist' and adjust the drep distr query

Sotatek-HuyLe3a:
#486 
- Add column 'registration_slot' to table 'drep_dist' and adjust the drep distr query
- Update protocol param after the hard fork init action is enacted

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepStatusProcessor.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepStatusProcessor.java
@@ -27,7 +27,6 @@ import static com.bloxbean.cardano.yaci.store.governanceaggr.GovernanceAggrConfi
 @RequiredArgsConstructor
 @EnableIf(value = STORE_GOVERNANCEAGGR_ENABLED, defaultValue = false)
 @Slf4j
-// todo: 'inactive', 'active again' case
 public class DRepStatusProcessor {
     private final DRepStorage dRepStorage;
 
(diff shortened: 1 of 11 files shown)


aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/DRepDistService.java
```diff
@@ -139,6 +166,21 @@ public void takeStakeSnapshot(int epoch) {
         end = System.currentTimeMillis();
         log.info(">> Indexes created for DRep dist temp tables <<" + (end - start) + " ms");
 
+        String excludeDelegationByRegistrationSlotCondition = isInBootstrapPhase
+                ? """
+            (
```
satran004, line 171: As we discovered, during the bootstrap phase, delegation to an unregistered dRep is possible. However, if the dRep remains unregistered during drep_distr, the delegation is ignored.

The condition below only checks whether delegation and drep_registration occur in the same slot. But what if the delegation happens in slot x and the registration occurs in slot x+5? Will this condition handle that case?

Also, how should records created during the bootstrap phase be handled once the phase is over? Does this issue mention about that?

https://github.com/IntersectMBO/cardano-ledger/issues/4598

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/DRepStatusProcessor.java
```diff
@@ -65,20 +64,25 @@ public void handleCommitEventToProcessDRepStatus(PreCommitEvent preCommitEvent)
...
-                            var regisDRep = dReps.stream().filter(dRep -> dRep.getDrepId().equals(drepRegistration.getDrepId())).findFirst();
+                        if (recentDRepRegistrationOpt.isEmpty()) {
+                            var regisDRep = dReps.stream()
+                                    .filter(dRep -> dRep.getDrepId().equals(drepRegistration.getDrepId()))
+                                    .reduce((first, second) -> second);
 
                             if (regisDRep.isEmpty()) {
                                 log.error("Cannot find recent dRep registration");
                             } else {
+                                dRepBuilder.registrationSlot(regisDRep.get().getRegistrationSlot());
```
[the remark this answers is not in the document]
Sotatek-HuyLe3a: >Will regisDRep.get().getRegistrationSlot() return the first registration entity

it will return the last registration entity in batch
the code intended to  handle the case : in that a batch can contain a registration entity, and we retrieve the last element, There may be cases iwhere in a batch, with the same drepId can have multiple registrations, deregistrations, and update....
```
    if (recentDRepRegistrationOpt.isEmpty()) {
                            var regisDRep = dReps.stream()
                                    .filter(dRep -> dRep.getDrepId().equals(drepRegistration.getDrepId()))
                                    .reduce((first, second) -> second);
```

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/DRepDistService.java
Sotatek-HuyLe3a: > Also, how should records created during the bootstrap phase be handled once the phase is over? Does this issue mention about that?
In the issue, there is mention of the need to clean up “dangling” delegations, but it does not clearly specify how to handle records created during the bootstrap phase once that phase ends, still needs to be looked into

satran004 (approved):
Thanks @Sotatek-HuyLe3a

PR #495, opened 2025-03-28
Add synchronized to fix connection issue during local query

Sotatek-HuyLe3a:
- Add synchronized to fix connection issue during local query
- Add Conway era checking during fetching data with local queries


satran004 (approved):
Thanks @Sotatek-HuyLe3a. Looks good

Issue #468, opened 2025-03-05
[governance] Add apis to return proposal status


satran004:
//TODO

PR #518, opened 2025-04-15
Add endpoints for Proposal, DRep data

Sotatek-HuyLe3a:
#468

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/api/governanceaggr/controller/DRepInfoController.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/api/governanceaggr/controller/DRepInfoController.java
@@ -16,7 +16,7 @@ import java.util.List;
 import static com.bloxbean.cardano.yaci.store.common.util.ControllerPageUtil.adjustPage;
 
 @RestController("DRepInfoController")
-@RequestMapping("${apiPrefix}/gov-aggr")
+@RequestMapping("${apiPrefix}/governance-state")
 @RequiredArgsConstructor
 @Slf4j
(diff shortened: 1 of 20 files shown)


aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/api/governanceaggr/controller/DRepInfoController.java
```diff
@@ -0,0 +1,42 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+import static com.bloxbean.cardano.yaci.store.common.util.ControllerPageUtil.adjustPage;
+
+@RestController("DRepInfoController")
+@RequestMapping("${apiPrefix}/gov-aggr")
```
satran004, line 19: @Sotatek-HuyLe3a Should we just use `governance` as path instead of `gov-aggr` as it's an internal name? What do you think?

satran004 (approved):

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/api/governanceaggr/controller/DRepInfoController.java
Sotatek-HuyLe3a: I changed from 'gov-aggr' to 'governance-state' @satran004

satran004: Thanks

PR #537, opened 2025-04-23
(DRep expiry) correct dormant epoch count calculation

Sotatek-HuyLe3a:


satran004 (approved):
This is a temporary fix, as the previous logic was completely wrong due to a typo. We need to reimplement the dRep expiry code based on our new findings and most likely add a new column, `active_until`, to the drep_dist table.

Issue #544, opened 2025-04-29
[Mainnet] Epoch Stake Amount Mismatch at Epoch 544


Sotatek-HuyLe3a:
## Summary
A discrepancy of **100B lovelace** was identified at epoch 544 for stake address  
`stake1uyguuqwdpexmhgjd07vax5t2ay3f7qvea49ex907g6fmvjclq6l03`.

- In **DB Sync**, the proposal refund was earned in **epoch 541** and reflected in the stake by epoch 544.
- In **Yaci Store**, the refund appears only from **epoch 544** onwards, resulting in a lower epoch stake amount.

## Root Cause (Preliminary)
The New Constitution proposal was likely **ratified at epoch 541** in DB Sync but treated as **expired at epoch 544** in Yaci Store, leading to different refund timings.

satran004:
This appears to be a regression. ProposalStatusProcessor uses `drepHash` for searches, whereas the existing repository implementation uses `drepId` column.​

https://github.com/bloxbean/yaci-store/blob/43629cf03b2a0bbd1174b2bad48f6bac5bb6fee4/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/storage/impl/repository/DRepDistRepository.java#L16

As a quick fix, we can convert drepHash to drepId before performing the search. Later, we can explore the possibility of searching by `drepHash` and `drepType`.

PR #546, opened 2025-05-01
Incorrect proposal status calculation in epoch 541 (mainnet) which causes missing proposal refund.

satran004:
#544 

- Search drep distr by drepId instead of drepHash
- New debug controller which is only enabled for debug profile


Sotatek-HuyLe3a (approved):

Sotatek-HuyLe3a:
Understand your changes, LGTM @satran004

PR #561, opened 2025-05-19
Introduce Content-Aware Rollback Events, Pre-Rollback Event, and Current Balance Tracking

satran004:
This PR introduces the following new features:

1. **Content-Aware Rollback Events**
2. **Pre-Rollback Event**
3. **Current Balance Tables for `address_balance` and `stake_address_balance`**

---

### 1. Content-Aware Rollback Events

Currently, our rollback strategy is based solely on the `slot` value. This works well for most tables, which are append-only and include a `slot` column.
However, this approach can cause issues for applications that maintain derived values using in-place updates.

To address this, we are introducing a new optional event type that informs applications of the specific data (e.g., addresses, UTxOs) that will be rolled back—*before* the rollback is applied.
This enables applications to adjust derived values or perform cleanup based on rollback impact.

#### In this PR, we introduce two content-aware rollback events:

* `UtxoRollbackEvent` in the UTXO store
* `BalanceRollbackEvent` in the Account store

`UtxoRollbackEvent` provides information about which `address_utxo` or `tx_input` entries will be deleted. Applications can use this to update application-specific values.

`BalanceRollbackEvent` is used in the Account store to handle rollback for the newly introduced "current" tables. These tables—`address_balance_current` and `stake_address_balance_current`—store the latest balances. See section 3 for details.

---

### 2. `PreRollbackEvent`

This is a new event that is published **just before** every `RollbackEvent`.
It provides applications an opportunity to inspect the current data before the actual rollback is applied to the core tables.
This allows applications to capture state, make adjustments, or log relevant data as needed.

---

### 3. New "Current" Balance Tables

Two new tables are introduced:

* `address_balance_current`
* `stake_address_balance_current`

These tables maintain the latest balance information for addresses and stake addresses. They are updated in sync with the main `address_balance` and `stake_address_balance` tables.

During a rollback, the system uses `BalanceRollbackEvent` to determine the affected address/unit pairs and update the corresponding rows in these current tables.

These current tables are especially useful for analytics queries, such as retrieving the top 100 addresses by balance.

---

### New Configuration Flags

These features are **optional**. Only `PreRollbackEvent` is always published by default.
The **Content-Aware Rollback Events** and **Current Balance Tables** are disabled by default and can be enabled using the following configuration flags:

#### For Account Store:

```properties
store.account.content-aware-rollback=true
store.account.current-balance-enabled=true
```

#### For UTXO Store:

```properties
store.utxo.content-aware-rollback=true
```

### DIFF
--- a/aggregates/account/build.gradle
+++ b/aggregates/account/build.gradle
@@ -4,6 +4,12 @@ dependencies {
     implementation(libs.parallel.collector)
 
     api("org.springframework.boot:spring-boot-starter-batch")
+
+    testImplementation project(':starters:spring-boot-starter')
+    testImplementation project(':starters:account-spring-boot-starter')
+    testImplementation("org.springframework.boot:spring-boot-starter-jooq")
+
+    testImplementation("org.flywaydb:flyway-core")
 }
 
 publishing {
--- a/aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/AccountStoreProperties.java
+++ b/aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/account/AccountStoreProperties.java
@@ -54,4 +54,10 @@ public class AccountStoreProperties {
 
     @Builder.Default
     private int pruningInterval = 86400;
+
+    @Builder.Default
+    private boolean contentAwareRollback = false;
+
+    @Builder.Default
+    private boolean currentBalanceEnabled = false;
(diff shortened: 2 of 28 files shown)


Sotatek-HuyLe3a:
Look good to me @satran004

Sotatek-HuyLe3a (approved):

PR #569, opened 2025-06-05
Support epoch stake pruning

Sotatek-HuyLe3a:
#493

### DIFF
--- a/aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/service/EpochStakePruningService.java
+++ b/aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/service/EpochStakePruningService.java
@@ -53,6 +53,10 @@ public class EpochStakePruningService {
 
     @Scheduled(fixedRateString = "${store.adapot.epoch-stake-pruning-interval:86400}", timeUnit = TimeUnit.SECONDS, initialDelay = 300)
     public void handleScheduledPruning() {
+        if (!adaPotProperties.isEpochStakePruningEnabled()) {
+            return;
+        }
+
         if (isPruning.get()) {
             log.debug("Epoch stake pruning is already in progress. Skipping this run !!!");
             return;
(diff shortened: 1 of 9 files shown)


aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/service/EpochStakePruningService.java
```diff
@@ -0,0 +1,119 @@
...
+    @Transactional
+    @Order(1)
+    public void handlePreAdaPotJobProcessingEvent(PreAdaPotJobProcessingEvent event) {
+        if (isPruning.get()) {
+            log.debug("Epoch stake pruning is already in progress. Skipping this run !!!");
+            return;
+        }
+
+        Thread.startVirtualThread(this::pruneOldEpochStakes);
+        System.out.println("Epoch stake pruning triggered by pre adapot job processing event");
```
satran004, line 74: Please change the sys out to log.

aggregates/adapot/src/main/java/com/bloxbean/cardano/yaci/store/adapot/service/EpochStakePruningService.java
```diff
@@ -0,0 +1,119 @@
...
+    @PostConstruct
+    public void init() {
+        log.info("<< Epoch Stake Pruning Service Enabled >>");
+        transactionTemplate = new TransactionTemplate(transactionManager);
+        transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+    }
+
+    @Scheduled(fixedRateString = "${store.adapot.epoch-stake-pruning-interval:86400}", timeUnit = TimeUnit.SECONDS, initialDelay = 300)
+    public void handleScheduledPruning() {
+        if (isPruning.get()) {
```
satran004, line 56: I think we can check the `store.adapot.epoch-stake-pruning-enabled` flag again in the scheduler method for the native image, since the ConditionalOnProperty annotation is evaluated at build time.

For the native image build, we can enable pruning during build time, so that the EpochStakePruningService bean is available. Then, in the `handleScheduledPruning` method, we can add an if-condition to check if pruning is enabled and proceed accordingly.

Sotatek-HuyLe3a: Done! I've added the check in handleScheduledPruning

satran004:
Thanks @Sotatek-HuyLe3a

satran004 (approved):

PR #578, opened 2025-06-13
New flag to enable/disable transaction witness storage

Sotatek-HuyLe3a:


Sotatek-HuyLe3a:
Thank @satran004 , 
 Agree with you, I think we should align with the existing pattern. need change name and default value.

PR #584, opened 2025-06-19
feat: Continue Sync on Parse Error

satran004:
### New Feature: Continue Sync on Parse Error

* Introduced a new optional flag: `store.continue-on-parse-error` (default: `false`).

  When set to `true`, the sync process will continue even if a parsing error occurs. By default, this flag is `false` to preserve current behavior. This option is intended for critical applications where uninterrupted syncing is preferred, even in the presence of rare parsing errors.

* If this flag is enabled and a parsing error occurs:

  * The error details will be saved in the `ERROR` table.
  * Sync will continue after logging the error, allowing users to investigate later while maintaining sync continuity.


satran004:
@matiwinnetou fyi.

Sotatek-HuyLe3a (approved):

PR #585, opened 2025-06-20
Simplify and consolidate database migration scripts , add txIndex field to voting procedures, gov action proposal -related classes

Sotatek-HuyLe3a:
- Removed redundant migration files and merged their changes into the initialization script to streamline database setup. This reduces duplication, improves maintainability, and ensures consistency across schema definitions.
- Add tx index to table 'voting_procedure' and 'gov_action_proposal' and update related classes. #571


components/events/src/main/java/com/bloxbean/cardano/yaci/store/events/domain/TxGovernance.java
```diff
@@ -15,6 +15,7 @@
 @Builder
 public class TxGovernance {
     String txHash;
```
satran004, line 17: I think this is a typo. We should also change txHash variable to private

satran004 (approved):

PR #587, opened 2025-06-20
Add methods to exclude inactive DReps from stake calculations and update accordingly in proposal status calculation

Sotatek-HuyLe3a:
#586


satran004 (approved):
Thanks @Sotatek-HuyLe3a

PR #605, opened 2025-07-02
chore: possible enhancement

nemo83:
# Ensure utxo cache populated before unspent are saved

The `UtxoStorageImpl` has an internal cache containing unspent outputs which is populated after spent inputs have been already processed.

This works well, but prevents classes extending the `UtxoStorageImpl` to build reliable filtering criteria when saving spent inputs, such has only save inputs for addresses of interest.

A bad implementation which doesn't cater for mempool tx (same-block chained tx) is:

```java
   @Override
    public void saveSpent(List<TxInput> txInputs) {
        var automaticPaymentInputs = txInputs
                .stream()
                .filter(txInput -> utxoRepository.findById(new UtxoId(txInput.getTxHash(), txInput.getOutputIndex())).isPresent())
                .toList();
        super.saveSpent(automaticPaymentInputs);
    }
```

If the cache was built before, one could check the cache first, fallback on the query, and if really missing, skip, or perform custom checks otherwise.

### DIFF
--- a/stores/utxo/src/main/java/com/bloxbean/cardano/yaci/store/utxo/processor/UtxoProcessor.java
+++ b/stores/utxo/src/main/java/com/bloxbean/cardano/yaci/store/utxo/processor/UtxoProcessor.java
@@ -35,8 +35,8 @@ import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.math.BigInteger;
-import java.util.Collections;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -59,7 +59,7 @@ public class UtxoProcessor {
     private final MeterRegistry meterRegistry;
 
     //List of utxos with pointer address. We need to fetch stake address for these in CommitEvent
-    private List<Utxo> utxosWithPointerAddress = Collections.synchronizedList(new ArrayList<>());
+    private final List<Utxo> utxosWithPointerAddress = Collections.synchronizedList(new ArrayList<>());
 
     @EventListener
     @Order(2)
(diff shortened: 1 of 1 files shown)


satran004:
@nemo83 Thanks.

@Sotatek-HuyLe3a could you please also have a look?

The change looks good to me. As you mentioned, it will help find the UTXO for an input created in the same block. But, this approach works reliably only when the sync is already at the tip and we are processing one block at a time.

During the initial sync, when we process blocks in parallel (e.g., batches of 100), there may be cases where the corresponding output for an input hasn’t been processed yet. In such scenarios, the input’s UTXO may not be found.

One option to handle this is by listening to the `CommitEvent`, which is published sequentially at the end of each batch. 

I am adding a few  possible solutions in the scope indexer here. There could also be a better solution, so we need to explore this further.

**Approach 1: Temporary input tracking**

1. In the `saveSpent()` method, store a list of inputs (tx hash, output index) in a temporary list.
2. In the `CommitEvent` handler, verify whether each input exists in the `address_utxo` table and delete any that don't.

**Approach 2: Deferred input saving**

* Skip saving the inputs in step 1, and instead, save only those inputs in step 2 after filtering for valid ones.

**Approach 3: Scheduled cleanup**

* Continue saving extra `tx_inputs` as-is. Periodically clean up invalid inputs (i.e., those whose outputs don’t exist) using a scheduled/bacground job, say every *n* number of blocks.

**Approach 4: Post-tip cleanup**

* During initial sync, save inputs as-is (including extras). Once sync reaches the tip, remove the invalid inputs from the `tx_inputs` table, and then switch to your proposed logic that checks for the existence of the UTXO during each block's processing.

In summary, this change is effective when syncing at the tip.
But, we also need to find best practices to handle inputs reliably during the initial parallel sync in scope indexer, without sacrificing performance or the benefits of batch processing.

satran004:
@nemo83 
>> One thing I never understood is why there's no foreign key on TxInput vs TxOutput to ensure integrity?

Avoiding foreign keys is one of the design guidelines in YS :)
Also, during parallel batch processing, it can cause issues since we don’t know which will be inserted first. Tx output or Tx input.
So, it’s an eventually consistent system.

PR #607, opened 2025-07-03
fix: add ReadOnly annotation to AccountBalanceProcessor, AccountBatchConfiguration

Sotatek-HuyLe3a:
Fix the bug that occurs when read-only is set to true


satran004 (approved):
Thanks @Sotatek-HuyLe3a . Looks good to me.

Issue #263, opened 2024-04-25
When receiving data from Cardano node, disconnection occurs intermittently.


MarMarMot:
I am building the Cardano node from the source code:
https://github.com/input-output-hk/cardano-node-wiki/blob/main/docs/getting-started/install.md
and run it in "Preprod" environment. 

In the last few days when I try to get data from the node, I get an error periodically. Each next attempt to retrieve data returns the same error, despite the message “Connection established”.

```
2024-04-25T10:55:37.362Z  INFO 114 --- [ntLoopGroup-5-3] c.b.c.y.s.core.service.CursorService     : # of blocks written: 1
2024-04-25T10:55:37.362Z  INFO 114 --- [ntLoopGroup-5-3] c.b.c.y.s.core.service.CursorService     : Block No: 169863
2024-04-25T10:55:37.374Z  WARN 114 --- [ntLoopGroup-5-3] c.b.c.y.s.a.p.AccountBalanceProcessor    : Account balance calculation will be ignored. Please run the aggregation app to calculate the account balance for pending blocks.
2024-04-25T10:55:37.374Z  WARN 114 --- [ntLoopGroup-5-3] c.b.c.y.s.a.p.AccountBalanceProcessor    : The last processed block for account balance calculation is not the same as the expected last block.
2024-04-25T10:55:37.374Z  WARN 114 --- [ntLoopGroup-5-3] c.b.c.y.s.a.p.AccountBalanceProcessor    : Last processed block for account balance: 15640
2024-04-25T10:55:37.374Z  WARN 114 --- [ntLoopGroup-5-3] c.b.c.y.s.a.p.AccountBalanceProcessor    : Current block: 169864
2024-04-25T10:55:37.375Z  INFO 114 --- [ntLoopGroup-5-3] c.b.c.y.s.core.service.CursorService     : # of blocks written: 1
2024-04-25T10:55:37.375Z  INFO 114 --- [ntLoopGroup-5-3] c.b.c.y.s.core.service.CursorService     : Block No: 169864
2024-04-25T10:57:09.378Z  WARN 114 --- [ntLoopGroup-4-3] i.netty.channel.DefaultChannelPipeline   : An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.

java.net.SocketException: Connection reset
        at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:401) ~[na:na]
[...]

PR #625, opened 2025-07-28
Fix: #263 Remove TipFinder bean. Create and destroy tip finder connection in find method

satran004:
#263

### Issue

To find the tip during startup, Yaci Store uses `TipFinderService`, which maintains a bean instance of `TipFinder`. This connection is kept alive solely for retrieving the tip, even though we only need it during startup or occasionally when reading local data directly from the node.

Since `TipFinder` does not have a KeepAlive agent configured, the connection gets reset over time, and `TipFinder` keeps trying to reconnect unnecessarily.

### Fix

Instead of keeping a persistent instance, we now create a new `TipFinder` instance for every `findTip` call and shut it down immediately after receiving the response. This avoids maintaining an unnecessary long-lived connection and simplifies resource usage.


Sotatek-HuyLe3a (approved):

nemo83:
OMG LET'S GOOOO

Issue #642, opened 2025-08-07
[preview] (epoch 993) Missing reward rest with address stake_test1up8xg8ur80nsymvefrlk4dshjyfrze4pep843l289f4n8fgrtrmy4


Sotatek-HuyLe3a:
When compared with DB-Sync, earned epoch = 993, the Yaci indexer is missing the reward_rest entries for the address stake_test1up8xg8ur80nsymvefrlk4dshjyfrze4pep843l289f4n8fgrtrmy4.

Cause: Some no-confidence proposals have the return address stake_test1up8xg8ur80nsymvefrlk4dshjyfrze4pep843l289f4n8fgrtrmy4.
In epoch 993, an update_committee proposal was ratified. According to the rules, those no-confidence proposals should be dropped in epoch 994.

Issue #618, opened 2025-07-14
[Preview] Some mismatches in drep stake distribution (compare to cardano db sync)


Sotatek-HuyLe3a:
##### DRep distribution mismatches

| Epoch(s)   | DRep ID                                                    | DRep Hash                                                | Note            |
|:-----------|:-----------------------------------------------------------|:---------------------------------------------------------|:----------------|
| 736        | drep1y285lm709qqh54a5z5t6vl2kaaxqmszpsxs36dgh3h2n7nqltj46y | 8f4fefcf28017a57b41517a67d56ef4c0dc04181a11d35178dd53f4c | Amount mismatch (vs DB Sync) |
| 730–731    | drep1ygsusq7j0wx67tzcz9u50rh9l2w6pmmsqglpzzqathlf0rqm87es5 | 21c803d27b8daf2c581179478ee5fa9da0ef70023e11081d5dfe978c | Amount mismatch (vs DB Sync) |
| 685–757    | drep1ytax4rwzvdwamnu67j2uk9z00660lpzcvmlys6266l9kt5c262k7m | fa6a8dc2635dddcf9af495cb144f7eb4ff845866fe48695ad7cb65d3 | Amount mismatch (vs DB Sync) |
| 682–684    | drep1ytax4rwzvdwamnu67j2uk9z00660lpzcvmlys6266l9kt5c262k7m | fa6a8dc2635dddcf9af495cb144f7eb4ff845866fe48695ad7cb65d3 | Only in Yaci Store (not in Db Sync) |
| 681–683    | drep1ytcw6qzpqqclx2yd0zy64ztvlkkhnf6yrzza8whgnq4vz5gh89626 | f0ed00410031f3288d7889aa896cfdad79a7441885d3bae8982ac151 | Only in Yaci Store (not in Db Sync) |
| 743–762    | drep1a2r7cf9jfyrqcp2dahfw2aw5lwfd3tymvdc8y5v0ad2652mfcw7   | ea87ec24b249060c054dedd2e575d4fb92d8ac9b637072518feb55aa | Found in DB Sync but not in Yaci Store |
| 709–830    | drep1ytkdk7j6g4jpnuamve0uy9kv2fyvlprdauv0ymtdpvkrt2gla0chu | ecdb7a5a456419f3bb665fc216cc5248cf846def18f26d6d0b2c35a9 | Amount mismatch (vs DB Sync) |

Sotatek-HuyLe3a:
[...]

PR #643, opened 2025-08-11
Fix: Correct proposal dropping logic for ratified/expired proposals

Sotatek-HuyLe3a:
## 📌 Purpose

This pull request aims to fix issues related to the calculation of DRep distribution and proposal dropping logic.

### Core Fix
- Updated the logic for dropping proposals: if a proposal is ratified, its sibling proposals and the descendants of those siblings will be dropped. 
- Added a temporary table `ss_gov_scheduled_to_drop_proposal_deposits` to store deposits of proposals scheduled to be dropped. These deposit values will be subtracted when calculating DRep voting power.

### Proposal Purpose-Based Dropping Logic
- **Purpose groups**: Proposals are grouped by purpose (PParamUpdatePurpose, HardForkPurpose, CommitteePurpose, ConstitutionPurpose)
- **Sibling identification**: Proposals with the same purpose and same parent are considered siblings
- **Dropping rules**: When a proposal is ratified, all its siblings and descendants of those siblings are dropped. When a proposal is expired, its descendants are dropped
- **Cross-purpose isolation**: Proposals of different purposes don't affect each other (e.g., ParameterChange doesn't drop HardFork proposals)
## 📎 Related Issues
#642, #618

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/ProposalRefundProcessor.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/processor/ProposalRefundProcessor.java
@@ -75,7 +75,7 @@ public class ProposalRefundProcessor {
         );
 
         ratifiedProposalsInThisEpoch.forEach(proposal ->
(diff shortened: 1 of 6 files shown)


satran004 (approved):
Thanks @Sotatek-HuyLe3a

Issue #685, opened 2025-09-29
Add conversion util for gov action id → tx hash/index


Sotatek-HuyLe3a:
## Description
CCL 0.7.0 introduces CIP-129 compatible `gov_action_id`, but does not yet provide a conversion to `(tx_hash, index)`.  
We should add a temporary utility in Yaci Store for this purpose.

## Proposal
- Create a util to map `gov_action_id → (tx_hash, index)`.
- Keep it at the API/user layer for now.
- Once CCL supports this, migrate from YS → CCL.

## Context
- CIP-129 standard adoption.
- Needed for other tools.

## Action items
1. Implement util in YS.
2. Use it in relevant APIs
3. Later: create CCL issue for native support.

PR #701, opened 2025-10-01
feat: add gov util support for cip-129

ducpm2303:
Add conversion util for gov action id → tx hash/index
https://github.com/bloxbean/yaci-store/issues/685

### DIFF
--- a/components/common/src/test/java/com/bloxbean/cardano/yaci/store/common/util/GovUtilTest.java
+++ b/components/common/src/test/java/com/bloxbean/cardano/yaci/store/common/util/GovUtilTest.java
@@ -189,7 +189,7 @@ class GovUtilTest {
     @DisplayName("Should handle maximum single-byte index value")
     void shouldHandleMaximumSingleByteIndexValue() {
         String txHash = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
-        int maxIndex = 290;
+        int maxIndex = 255;
         
         String govActionId = GovId.govAction(txHash, maxIndex);
         GovActionId parts = GovUtil.toGovActionIdFromBech32(govActionId);


components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,125 @@
...
+public class GovUtil {
+
+    /**
+     * Converts a CIP-129 compatible gov_action_id bech32 string to transaction hash and index.
+     * 
+     * @param govActionIdBech32 The gov_action_id in bech32 format (e.g., "gov_action1...")
+     * @return GovActionIdParts containing txHash and index
+     * @throws IllegalArgumentException if the input is invalid
+     */
+    public static GovActionIdParts convertGovActionIdToTxHashAndIndex(String govActionIdBech32) {
```
satran004, line 28: Should we rename it to `govActionIdToTxHashAndIndex` ? 
@Sotatek-HuyLe3a Do you have a better suggestion for this method name?

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Creates a CIP-129 compatible gov_action_id from transaction hash and index.
+     * This method uses CCL's native implementation.
+     * 
+     * @param txHash The transaction hash (32-byte hex string)
+     * @param index The governance action index
+     * @return CIP-129 compliant gov_action_id
+     * @throws IllegalArgumentException if the input is invalid
+     */
+    public static String createGovActionId(String txHash, long index) {
```
satran004, line 65: @ducpm2303 In Cardano Client Lib, we already have a method for this. You can check GovId class. So we probably don't need it here.

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,125 @@
...
+
+        try {
+            return GovId.govAction(txHash, (int) index);
+        } catch (Exception e) {
+            log.error("Failed to create gov_action_id from txHash: {} and index: {}", txHash, index, e);
+            throw new IllegalArgumentException("Failed to create gov_action_id: " + e.getMessage(), e);
+        }
+    }
+
+    public static class GovActionIdParts {
```
satran004, line 86: Can we convert it to record?

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+@Slf4j
+@UtilityClass
+public class GovUtil {
+
+    /**
+     * Converts a CIP-129 compatible gov_action_id bech32 string to transaction hash and index.
+     * 
+     * @param govActionIdBech32 The gov_action_id in bech32 format (e.g., "gov_action1...")
+     * @return GovActionIdParts containing txHash and index
```
satran004, line 25: Let's add a //TODO in the comment. This is temporary implementation, it will be moved to CCL.

satran004 (commented):
Thanks @ducpm2303  . Few minor comments

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
Sotatek-HuyLe3a: It might be better if we rename it to toGovActionId, and instead of returning GovActionIdParts, we return GovActionId (the class in transaction-spec in CCL). This would be somewhat similar to toDrep in the GovId class. Not sure if this is necessarily better, but I think it could make things more consistent.
https://github.com/bloxbean/cardano-client-lib/blob/master/governance/src/main/java/com/bloxbean/cardano/client/governance/GovId.java#L102C24-L102C30

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,65 @@
...
+public class GovUtil {
+
+    /**
+     * Converts a CIP-129 compatible gov_action_id bech32 string to transaction hash and index.
+     * //TODO: This is temporary implementation, it will be moved to CCL.
+     * @param govActionIdBech32 The gov_action_id in bech32 format (e.g., "gov_action1...")
+     * @return GovActionIdParts containing txHash and index
+     * @throws IllegalArgumentException if the input is invalid
+     */
+    public static GovActionIdParts toGovActionId(String govActionIdBech32) {
```
satran004, line 28: @ducpm2303 Let's change the return type to `GovActionId` which is part of CCL.

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,65 @@
...
+        if (govActionIdBech32 == null || govActionIdBech32.trim().isEmpty()) {
+            throw new IllegalArgumentException("gov_action_id cannot be null or empty");
+        }
+
+        try {
+            if (!govActionIdBech32.startsWith("gov_action1")) {
```
satran004, line 34: Should we check against `gov_action` as that's the exact prefix?

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/util/GovUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+            byte[] decodedBytes = Bech32.decode(govActionIdBech32).data;
+            
+            if (decodedBytes.length != 33) {
+                throw new IllegalArgumentException("Invalid gov_action_id: expected 33 bytes, got " + decodedBytes.length);
+            }
+
+            byte[] txHashBytes = Arrays.copyOfRange(decodedBytes, 0, 32);
+            String txHash = HexUtil.encodeHexString(txHashBytes);
+            int index = decodedBytes[32] & 0xFF;
```
satran004, line 46: Is it safe to assume that index is 1 byte only? Looks like spec doesn't specify that explicitely.

components/common/src/test/java/com/bloxbean/cardano/yaci/store/common/util/GovUtilTest.java
```diff
@@ -0,0 +1,268 @@
...
+        // ISSUE: CCL GovId.govAction() has a bug with indices > 255
+        // 
+        // Problem Analysis:
+        // 1. CCL GovId.govAction() incorrectly concatenates the index to the txHash as a hex string
+        // 2. Example: txHash + Integer.toHexString(index) = "abc...def100" for index=256
+        // 3. This creates an invalid hex string with odd length (not divisible by 2)
+        // 4. HexUtil.decodeHexString() throws IllegalArgumentException for invalid hex strings
+        // 
+        // Root Cause:
+        // The CCL implementation doesn't follow CIP-129 spec which requires:
```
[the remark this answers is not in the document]
satran004: @ducpm2303 Can you create an issue in CCL repo and later you can create a PR to fix that?

This PR looks ok to me. Any change still pending?

satran004 (approved):

Issue #711, opened 2025-10-31
Add support for `tx-hash -> tx-cbor` queries


paolino:
In the context of MPFS we are experimenting to improve the trustless public setup.

MPFS has some end-points producing unsigned transaction and we want to augment those responses with a context to resolve the transaction inputs. One valid context would be the pre-images of all the transaction input hashes. 

With that addition the client will be able to compute the UTxO consumed by the transaction. This will greatly improve the power of the pre-signature vetting phase which is our main goal for the client to trust a public MPFS.

As yaci-store is seeing those pre-images as transaction cbors at the N2N protocol level, we would like those to be stored along with their hashes and efficiently served by hash.

As we run the yaci-store db in our setup we are happy also to query the directly table behind the scenes, in case this would improve the chances of this issue to be tackled. 

BTW, thanks for this great indexer!

PR #713, opened 2025-11-04
feat: implement CBOR storage and retrieval for blocks and transactions

ducpm2303:
https://github.com/bloxbean/yaci-store/issues/711
- Added BlockCborStorage and TransactionCborStorage interfaces for managing CBOR data.
- Implemented BlockCborStorageImpl and TransactionCborStorageImpl for saving and deleting CBOR data.
- Introduced BlockCborPruningService and TransactionCborPruningService for automatic pruning of old CBOR data based on retention settings.
- Created REST APIs for retrieving block and transaction CBOR data, including existence checks.
- Added necessary database migrations for CBOR tables in H2, MySQL, and PostgreSQL.
- Enhanced Block and Txn domain models to include CBOR byte arrays.

### DIFF
--- a/stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/BlocksStoreProperties.java
+++ b/stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/BlocksStoreProperties.java
@@ -33,7 +33,7 @@ public class BlocksStoreProperties {
 
     /**
      * Retention period for CBOR data in slots.
-     * Default: 43,200 slots (~5 days based on 2160 safe blocks).
+     * Default: 43,200 slots (based on 2160 safe blocks).
      * CBOR data older than this will be pruned if cborPruningEnabled is true.
      */
     @Builder.Default
--- a/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
+++ b/stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
@@ -46,7 +46,7 @@ public class TransactionStoreProperties {
 
     /**
      * Retention period for CBOR data in slots.
-     * Default: 43,200 slots (~5 days based on 2160 safe blocks).
+     * Default: 43,200 slots (based on 2160 safe blocks).
      * CBOR data older than this will be pruned if cborPruningEnabled is true.
      */
     @Builder.Default
--- a/starters/blocks-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/blocks/BlocksStoreAutoConfigProperties.java
+++ b/starters/blocks-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/blocks/BlocksStoreAutoConfigProperties.java
@@ -36,7 +36,7 @@ public class BlocksStoreAutoConfigProperties {
 
        /**
         * Retention period for CBOR data in slots.
-        * Default: 43,200 slots (~5 days based on 2160 safe blocks).
+        * Default: 43,200 slots (based on 2160 safe blocks).
         * CBOR data older than this will be pruned if cborPruningEnabled is true.
         */
        private int cborRetentionSlots = 43200; // 20 * 2160 slots
--- a/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionAutoConfigProperties.java
+++ b/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionAutoConfigProperties.java
@@ -52,7 +52,7 @@ public class TransactionAutoConfigProperties {
 
       /**
        * Retention period for CBOR data in slots.
-       * Default: 43,200 slots (~5 days based on 2160 safe blocks).
+       * Default: 43,200 slots (based on 2160 safe blocks).
        * CBOR data older than this will be pruned if cborPruningEnabled is true.
        */
       private int cborRetentionSlots = 43200; // 20 * 2160 slots


stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/storage/BlockCborStorage.java
```diff
@@ -0,0 +1,26 @@
+package com.bloxbean.cardano.yaci.store.blocks.storage;
+
+import com.bloxbean.cardano.yaci.store.blocks.domain.Block;
+
+/**
+ * Storage interface for block CBOR data
+ */
+public interface BlockCborStorage {
```
satran004: Thanks @ducpm2303 for the PR.

1. Instead of introducing a separate BlockCborStorage, can we have an additional method (saveCbor) in existing BlockStorage api?

The BlockStorageImpl will have both BlockRepository and BlockCborRepository. From, processor layer , then it can invoke `saveCbor` method.

But we can keep the same `deleteBySlotGreaterThan` method and invoke both blockRepository.delete and blockCborRepository.delete from that.

Same pattern can be followed in TransactionStorage,

2. I am not sure about the controller though. Because with separate controllers, we can enable/disable controllers with config. Alternatively, to keep it simple, we can have just one extra api endpoint in the existing controller, which can return 404 or some other error like `Unsupported` when flag is disabled.

3. Block/Txn domain classes: I am not sure about adding cbor field to existing domain class like Block and Txn. Most of the time caller requesting for block may not need this data. So probably keeping separate domain class may make sense here. BlockCbor.

stores-api/transaction-api/src/main/java/com/bloxbean/cardano/yaci/store/api/transaction/controller/TransactionController.java
```diff
@@ -66,4 +74,59 @@ public TransactionPage getTransactions(@RequestParam(name = "page", defaultValue
     public List<TxnWitness> getTransactionWitnesses(@PathVariable @Pattern(regexp = "^[0-9a-fA-F]{64}$") String txHash) {
         return transactionService.getTransactionWitnesses(txHash);
     }
+    
+    @GetMapping(value = "{txHash}/cbor", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
```
satran004: Should we also enable these endpoints to return cbor hex string? A separate endpoint is ok. 
Alternative, if possible, if client sends Accept header as `application\json`, we can wrap the cbor hex string in a json and return 

`{ "cborHex": "84a200581c1c7b4a9e..." }`

satran004: Same for BlockController

stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/BlocksStoreProperties.java
```diff
@@ -14,4 +14,27 @@ public class BlocksStoreProperties {
...
+     */
+    @Builder.Default
+    private boolean cborPruningEnabled = false;
+
+    /**
+     * Retention period for CBOR data in slots (default: 129600 = ~30 days).
+     * CBOR data older than this will be pruned if cborPruningEnabled is true.
+     */
+    @Builder.Default
+    private int cborRetentionSlots = 129600; // 30 days worth of slots
```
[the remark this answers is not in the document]
ducpm2303: updated @Sotatek-HuyLe3a could you please confirm ?

stores/transaction/src/test/resources/application.yml
```diff
@@ -19,5 +19,9 @@ logging:
 store:
   cardano:
     protocol-magic: 764824073
+  transaction:
```
Sotatek-HuyLe3a: it might be better to keep transaction.save-cbor and blocks.save-cbor inside comments (or documented in the config) to make it clear that these flags are optional

stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/TransactionStoreProperties.java
```diff
@@ -27,4 +27,27 @@ public class TransactionStoreProperties {
...
+     */
+    @Builder.Default
+    private boolean cborPruningEnabled = false;
+
+    /**
+     * Retention period for CBOR data in slots (default: 129600 = ~30 days).
+     * CBOR data older than this will be pruned if cborPruningEnabled is true.
+     */
+    @Builder.Default
+    private int cborRetentionSlots = 129600; // 30 days worth of slots
```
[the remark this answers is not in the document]
Sotatek-HuyLe3a: In a typical scenario, the safe slot value could be 2160. However, I’m not fully confident about more complex contexts, especially when certain features are enabled/disabled, and whether deleting data could lead to any unexpected side effects.

Personally, I think that for the transaction table, we might safely set the safe slot to 43200 (similar to what is already used in the existing transaction pruning feature).

@ducpm2303 , could you help verify if there are any other features that still reference old transactions, or if any logic from other components depends on the transaction table? Personally, I believe there isn’t, but would like to double-check.

stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/BlocksStoreProperties.java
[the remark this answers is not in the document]
Sotatek-HuyLe3a: The number of safe block is 2160 (for rollback) -> safe slot might be 20 * 2160 = 43200. 
With the value 129600, I think it is not 30 days slot.

stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/storage/impl/BlockCborStorageImpl.java
```diff
@@ -0,0 +1,58 @@
...
+@Slf4j
+public class BlockCborStorageImpl implements BlockCborStorage {
+
+    private final BlockCborRepository blockCborRepository;
+    private final BlocksStoreProperties blocksStoreProperties;
+
+    @Override
+    @Transactional
+    public void save(BlockCbor blockCbor) {
+        if (!blocksStoreProperties.isSaveCbor()) {
```
satran004: @ducpm2303 I think we can keep this check in processor layer and keep storage layer simple.
So, if in the future, we introduce key value based blockCborStorage, we don't need to do any duplicate check. Same for delete method for rollback.

stores/blocks/src/main/java/com/bloxbean/cardano/yaci/store/blocks/processor/BlockProcessor.java
```diff
@@ -26,10 +30,16 @@
 @EnableIf(STORE_BLOCKS_ENABLED)
 @Slf4j
 public class BlockProcessor {
```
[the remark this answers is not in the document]
ducpm2303: @satran004 It seems ByronTx doesn’t have CBOR properties, and the yaci core logic doesn’t deserialize the cbor field either 
https://github.com/bloxbean/yaci/blob/main/core/src/main/java/com/bloxbean/cardano/yaci/core/model/serializers/ByronBlockSerializer.java#L203-L228

stores/transaction/src/test/resources/application.yml
Sotatek-HuyLe3a: if it's in 'test' folder. I think we can keep it. At first i thought it wasn't in the test folder

Sotatek-HuyLe3a (approved):

PR #738, opened 2025-11-18
 (governance-aggr-api): add 'DROPPED' proposal status, derive proposal status from AdaPot jobs

Sotatek-HuyLe3a:
- Replace EpochParamStorage with AdaPotJobStorage to base proposal status on the last completed AdaPot job epoch
  - Add DROPPED proposal status and update corresponding logic
  - Refactor proposal DTO construction into a method


satran004 (approved):
Thanks @Sotatek-HuyLe3a . Looks good

PR #739, opened 2025-11-18
Use latest block epoch for DRep API and keep inactive DRep voting power

Sotatek-HuyLe3a:


aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/storage/impl/DRepStorageReaderImpl.java
```diff
@@ -238,7 +238,7 @@ public Optional<DRepDetailsDto> getDRepDetailsByDRepId(String drepId, int epoch)
                 ? DRepStatus.ACTIVE
                 : DRepStatus.INACTIVE;
 
-        if (status == DRepStatus.INACTIVE || status == DRepStatus.RETIRED) {
```
satran004: Does this mean that an inactive dRep will also have voting power?

satran004: @Sotatek-HuyLe3a I think we can keep voting power as is. Amount may create more confusion.

Issue #797, opened 2026-02-09
[Blockforst] Add H2 Dialect Support for Existing Blockfrost Extension Modules (address, assets)


ducpm2303:
## Summary

The current blockfrost extension modules (`address`, `asset`) contain PostgreSQL-specific SQL queries that will fail when running against H2 (used for tests and lightweight deployments). We need to add H2 fallback queries so these modules can run in both environments.

The `epoch` module is already fully portable (type-safe JOOQ only) and requires no changes.

## Background

During the asset module implementation, we documented JOOQ best practices and a dialect-branching pattern in [blockfrost-architecture-options.md](extensions/blockfrost/blockfrost-architecture-options.md). The core codebase already has a reference implementation in `UtxoStorageReaderImpl.getUnitCondition()` that branches on `dsl.dialect().family() == SQLDialect.POSTGRES`.

H2 is used for **tests and lightweight development** with small datasets, so the fallback queries can be simpler and less optimized than the PostgreSQL versions.

### PostgreSQL-Specific Features Used

| Feature | H2 Equivalent | Used In |
|---------|---------------|---------|
| `DISTINCT ON (col)` | `GROUP BY col` + `MIN()` | asset |
| `MATERIALIZED` CTE hint | Omit keyword (plain subquery) | asset |
| `LATERAL` join | Correlated subquery or application-side join | address, asset |
| `::type` casting (`::jsonb`, `::bigint`, `::numeric`) | `CAST(x AS TYPE)` | address, asset |
| JSONB `@>` containment | `CAST(col AS VARCHAR) LIKE '%...%'` | asset |
| `jsonb_to_recordset()` | Parse JSON string in application code | address |
| `jsonb_array_elements()` | Parse JSON string in application code | asset |
| JSONB `->` / `->>` operators | Parse JSON string in application code | asset |

---

#### 1. `getAddressTotal(String address)`
- **PG features:** `jsonb_to_recordset({0}::jsonb)`, `DSL.lateral()`, `::jsonb` casting
- **H2 approach:** Fetch raw UTXO rows, parse `amounts` JSON column as string in application code, aggregate in-memory
- **Complexity:** Medium

#### 2. `findUnspentAddressBalanceByUnit(String address)`
- **PG features:** `jsonb_to_recordset({0}::jsonb)`, `DSL.lateral()`, `::jsonb` casting
- **H2 approach:** Same as above — fetch UTXOs, parse JSON in app, aggregate
- **Complexity:** Medium

[...]

PR #799, opened 2026-02-11
feat: support H2 querry for blockforst modules

ducpm2303:
https://github.com/bloxbean/yaci-store/issues/797

### DIFF
--- a/config/application.properties
+++ b/config/application.properties
@@ -344,12 +344,12 @@ logging.level.com.bloxbean.cardano.yaci.store.core.service=INFO
 
 #######################################################################################
 # Blockfrost Configuration
-# Default value = "true", To disable an endpoint, set the flag to "false"
+# Default value = "false", To enable an endpoint, set the flag to "true"
 #######################################################################################
 blockfrost.apiPrefix=/api/v1/blockfrost
-#store.extensions.blockfrost.epoch.enabled=true
-#store.extensions.blockfrost.address.enabled=true
-#store.extensions.blockfrost.asset.enabled=true
+#store.extensions.blockfrost.epoch.enabled=false
+#store.extensions.blockfrost.address.enabled=false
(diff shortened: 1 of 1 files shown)


extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/impl/BFAddressStorageReaderImpl.java
```diff
@@ -169,6 +171,13 @@ public Map<String, BigInteger> findUnspentAddressBalanceByUnit(String address) {
      * JSON expansion where possible, and JSON amounts are used for non-lovelace assets.
      */
     private Map<String, BigInteger> fetchAmountSums(Condition addressCondition, boolean spent) {
+        if (BlockfrostDialectUtil.isPostgres(dsl)) {
+            return fetchAmountSumsPostgres(addressCondition, spent);
+        }
+        return fetchAmountSumsH2(addressCondition, spent);
```
Sotatek-HuyLe3a: Can you also check with Mysql? If h2 query can still work on mysql, our condition should be  if(postgres) ... else .... h2/mysql

config/application.properties
```diff
@@ -344,12 +344,12 @@ logging.level.com.bloxbean.cardano.yaci.store.core.service=INFO
 
 #######################################################################################
 # Blockfrost Configuration
-# Default value = "false", To enable an endpoint, set the flag to "true"
+# Default value = "true", To disable an endpoint, set the flag to "false"
```
satran004: @ducpm2303 I think we need to make default value to false. Also in corresponding auto config properties file. So BF api will be an optional component.

satran004 (approved):

Issue #826, opened 2026-03-02
[analytics-store] Missing genesis data in Parquet export


Sotatek-HuyLe3a:
When exporting data from PostgreSQL to Parquet format in the analytics store, genesis data (slot=-1) is not being included in the output. This results in missing records that exist in Postgres but are absent from the generated Parquet files.

PR #837, opened 2026-03-05
fix: Include genesis data in Parquet exports for all networks

Sotatek-HuyLe3a:
#826 

##  Problem                                                                                                 

  Genesis records in PostgreSQL are stored with slot = -1. The analytics-store Parquet export was missing these records due to two bugs:

###  Bug 1 — Slot range excluded genesis on networks 

In Yaci-Store, the slot in genesis block = -1, so we need include "-1" in SlotRange.

 ### Bug 2 — Premature export during initial sync

 **GapDetectionService.getExportEndDate()** clamped the export end date to never be before genesis date. During initial sync, this bypassed the day buffer, for example, 2-day buffer:

  **latestSynced** = 2017-09-23 (only a few blocks synced) (2021-09-23 is genesis day in mainnet)
  **calculatedEndDate** = 2017-09-23 - 2 days = 2017-09-21
  clamp to genesis → 2017-09-23  ← buffer bypassed!
  → Export runs with only 1 block in DB → marks COMPLETED → data permanently lost

###  Changes

  **GapDetectionService.java** — Remove the genesis-date clamp in getExportEndDate(). When the buffer pushes the end date before genesis, findMissingExports() already returns an empty list (endDate < startDate check), correctly deferring export until sync has progressed past the buffer period.

  **PartitionValue.java** — Change slot clamping from Math.max(0, startSlot) to startSlot <= 0 ? -1 : startSlot. Any partition whose calculated start slot is ≤ 0 is the genesis day — set to -1 so WHERE slot >= -1 captures
  genesis records.

  **SlotRange.java** — Relax validation to accept startSlot >= -1 (was >= 0) to accommodate genesis data.

  Tests added:
  - **DatePartitionTest** — 5 tests covering mainnet, preprod, normal days, pre-genesis edge case, and path formatting
  - **SlotRangeTest** — 7 tests covering valid ranges, genesis start slot, slot count, and validation errors

### DIFF
--- a/aggregates/analytics-store/src/main/java/com/bloxbean/cardano/yaci/store/analytics/gap/GapDetectionService.java
+++ b/aggregates/analytics-store/src/main/java/com/bloxbean/cardano/yaci/store/analytics/gap/GapDetectionService.java
@@ -118,25 +118,14 @@ public class GapDetectionService {
      * This ensures we don't export data for dates that are still being populated,
      * providing a safety buffer during initial blockchain synchronization.
      *
-     * The returned date is clamped to never be before genesis, even if the buffer
-     * period extends before the blockchain's start date.
+     * If the buffer pushes the end date before genesis, return it as-is.
      *
-     * @return Safe end date for exports as LocalDate (>= genesis date)
+     * @return Safe end date for exports as LocalDate
      */
     public LocalDate getExportEndDate() {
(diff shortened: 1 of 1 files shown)


satran004 (approved):
LGTM. Thanks @Sotatek-HuyLe3a

PR #861, opened 2026-03-24
feat(core): move SyncStatusService from admin-ui to yaci-core

### LINKED ISSUE #860: Move SyncStatusService from admin-ui to yaci-core
## Summary

`SyncStatusService` currently lives in the `admin-ui` module, but it provides general-purpose sync status information (current block/slot/epoch, sync percentage, network tip) that is useful beyond the admin UI. It should be promoted to `yaci-core` alongside the existing `HealthService`.

## Motivation

- **`HealthService`** (in `core.service`) provides basic connection health (alive, receiving blocks, last block time) — sufficient for simple liveness checks.
- **`SyncStatusService`** (in `admin-ui`) provides richer sync progress information: current block/slot/epoch, network tip, sync percentage, era, and whether the node is fully synced. Many consumers need this without pulling in the entire admin-ui module.

Both services serve related but distinct needs — `HealthService` answers *"is it working?"* while `SyncStatusService` answers *"how far along is it?"*. Making `SyncStatusService` available in core allows any module or downstream project to query sync progress.

## Proposed Changes

- Introduce a `SyncStatus` domain object in `common.domain` (following the `HealthStatus` pattern)
- Move/recreate `SyncStatusService` in `core.service` (following the `HealthService` pattern)
- Deprecate and eventually remove `SyncStatusService` from `admin-ui`, delegating to the core implementation to maintain backwards compatibility during the transition
- Ensure the admin-ui API responses remain unchanged (backwards compatible)

## Additional Context

`HealthService` will remain as the lightweight option for consumers who only need connection health. `SyncStatusService` in core will serve consumers who need detailed sync progress (percentage, network tip comparison, etc.).

matiwinnetou:
## Summary
- Introduces `SyncStatus` record in `common.domain` and `SyncStatusService` in `core.service`, making sync progress (percentage, network tip, epoch, era) available to any module without depending on admin-ui
- Removes `SyncStatusService` and `SyncStatusDto` from admin-ui entirely — controller uses core's service directly
- `LedgerStateStatusService` now depends on core's `SyncStatusService`
- Sync control operations (start/stop/restart) inlined in the controller using `StartService` directly, removing the unnecessary service wrapper that mixed status retrieval with control concerns

Closes #860

## Test plan
- [x] Unit tests for `SyncStatusServiceTest` (9 tests covering: no cursor, synced/unsynced, Byron/null era, tip unavailable, epoch failure, tip caching)

### DIFF
--- a/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/SyncStatusService.java
+++ b/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/SyncStatusService.java
@@ -48,10 +48,12 @@ public class SyncStatusService {
             blockHash = cursor.getBlockHash();
             era = cursor.getEra() != null ? cursor.getEra().name() : "Unknown";
 
+            // Calculate current epoch from sync slot
             if (cursor.getEra() != null && cursor.getEra() != Era.Byron) {
                 try {
                     currentEpoch = eraService.getEpochNo(cursor.getEra(), currentSlot);
                 } catch (Exception e) {
+                    // Fall back to 0 if epoch calculation fails
                     log.debug("Could not calculate epoch from slot: {}", e.getMessage());
                 }
             }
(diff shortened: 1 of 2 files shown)


components/admin-ui/src/main/java/com/bloxbean/cardano/yaci/store/adminui/controller/SyncStatusController.java
```diff
@@ -18,11 +19,12 @@
 @Hidden
 public class SyncStatusController {
     private final SyncStatusService syncStatusService;
+    private final StartService startService;
```
satran004: @Sotatek-HuyLe3a Using `StartService` here should not be an issue, but can you check what will be the behavior for readOnly(true) mode?

PR #866, opened 2026-03-25
feat: implement Blockfrost Network API

### LINKED ISSUE #864: Create a Blockfrost compatible API for Governance

ducpm2303:
## Module: Blockfrost Network APIs (4 endpoints)

Closes #864

---

## Summary

This PR implements the Blockfrost network module end-to-end:

- controller, service, storage reader, DTOs, mapper, and starter wiring
- all 4 network endpoints under `/`, `/genesis`, `/network`, and `/network/eras`
- genesis parameters sourced from bundled `GenesisConfig` (no DB queries)
- era boundary computation using `EraService` with correct Byron/Shelley slot arithmetic
- `supply.circulating` computed as `UTxO + spendable reward + spendable reward_rest − withdrawals`, matching Blockfrost/cardano-db-sync formula (validated: ~4 ADA sync lag)

---

## Review guide

| Area | Files | What to review |
|---|---|---|
| Wiring | `BFNetworkConfiguration`, `BFAutoConfiguration`, `BFProperties`, starter/build files | module registration and feature-flag wiring |
| API surface | `BFNetworkController`, `BFNetworkService`, DTOs | endpoint contract and response shape |
| Mapping | `BFNetworkMapper`, `BFNetworkMapperDecorator` | adapot → BF supply/stake DTO mapping |
| Query logic | `BFNetworkStorageReaderImpl` | circulating supply formula, epoch lookup |
| Genesis/era logic | `BFNetworkService.getGenesis()`, `getNetworkEras()` | slot arithmetic, safe-zone constants, touching era boundaries |

### Data flow

```mermaid
flowchart LR
    A[BFNetworkController] --> B[BFNetworkService]
    B --> C[BFNetworkStorageReaderImpl]
    B --> D[BFNetworkMapper]
    B --> E[GenesisConfig]
    B --> F[EraService]

    C --> G[(address_utxo)]
    C --> H[(reward / reward_rest)]
    C --> I[(withdrawal)]
    C --> J[(adapot)]
    D --> J
```

---

## Endpoints

| # | Endpoint | Status | Notes |
|---|---|---|---|
| 1 | `GET /` | ⚠️ Static diff | `url` and `version` reflect local service identity, not Blockfrost hosted values — accepted |
| 2 | `GET /genesis` | ✅ Match | all 10 fields exact match on preprod |
| 3 | `GET /network` | ⚠️ Partial | `circulating` ✅ implemented; `locked` ⚠️ stub (accepted limitation); `live` ⚠️ structural gap (see accepted limitations) |
| 4 | `GET /network/eras` | ⚠️ Partial | eras 0–5 exact match; Conway `end` null — Protocol v11 hardfork not yet fired (scheduled epoch 285) |

---

### Main implementation

- added the full Blockfrost network module:
  - controller
  - service
  - DTOs (`BFRootDto`, `BFGenesisDto`, `BFNetworkDto`, `BFEraDto`)
  - mapper
  - storage reader
  - starter / autoconfiguration wiring

### Follow-up correctness fixes validated live

| Area | Fix |
|---|---|
| Era boundaries | touching boundary arithmetic: `end` of each era equals `start` of the next era; Byron uses absolute time from genesis start; post-Byron eras use `eraService.blockTime()` |
| `supply.circulating` | implemented as `SUM(unspent UTxO) + SUM(spendable reward) + SUM(spendable reward_rest) − SUM(withdrawals)`; ~4 ADA sync-lag diff from Blockfrost |

---

## Validation

Live validation was run against:

- source of truth: `https://cardano-preprod.blockfrost.io/api/v0`
- local target: `http://localhost:8088/api/v1/blockfrost`
- runtime: local `yaci-store` rebuilt from current source and redeployed as `bloxbean/yaci-store:dev`

### Result matrix

| Endpoint | Checks | Matched | Accepted | Notes |
|---|---:|---:|---:|---|
| `GET /` | 2 | 0 | 2 | url + version static identity diff |
| `GET /genesis` | 10 | 10 | 0 | exact match |
| `GET /network` | 9 | 6 | 3 | circulating ✅; locked stub; live structural gap |
| `GET /network/eras` | 61 | 60 | 1 | Conway `end` null — Protocol v11 hardfork not yet fired |
| **Total** | **82** | **76** | **6** | |

### Field-level detail: `GET /network`

| Field | BF (epoch 284) | Local | Status |
|---|---|---|---|
| `supply.max` | `"45000000000000000"` | `"45000000000000000"` | ✅ |
| `supply.total` | `"31867356966492467"` | `"31867356966492467"` | ✅ |
| `supply.treasury` | `"1813456630289959"` | `"1813456630289959"` | ✅ |
| `supply.reserves` | `"13132643033507533"` | `"13132643033507533"` | ✅ |
| `supply.circulating` | `"30053068171434944"` | `"30053068175836863"` | ✅ ~4 ADA sync lag |
| `supply.locked` | `"173187540647343"` | `"0"` | ⚠️ stub |
| `stake.active` | `"552991509235407"` | `"552991509235407"` | ✅ |
| `stake.live` | `"488812434455346"` | `"0"` | ⚠️ structural gap |

---

## Required configuration

| Flag | Required for |
|---|---|
| `store.extensions.blockfrost.network.enabled=true` | network module endpoints |
| `store.adapot.enabled=true` | `supply.total`, `treasury`, `reserves`, `stake.active` |
| `store.adapot.api-enabled=true` | `NetworkInfoApiService` bean required by `BFNetworkService` |

---

### `GET /` — static identity fields

`url` and `version` reflect the local service identity, not the hosted Blockfrost service. Expected for any self-hosted deployment.

### `GET /network` — `supply.locked` stub

`locked` (UTxO at script addresses) returns `"0"`. The formula is known and was verified against the DB (computed `173,149,370,262,540` vs BF `173,149,725,609,652` — ~355 ADA sync lag), but this field is accepted as a permanent stub in this PR.

### `GET /network` — `stake.live` structural gap

`live` stake returns `"0"`. Blockfrost sources this from the Cardano node's in-memory mark snapshot via the Local State Query protocol (`QueryStakeSnapshot`). Every DB-table-based approach overshoots by 43–101B because the node's incremental stake distribution is updated block-by-block in memory and is not available in yaci-store's on-chain tables. This is a permanent limitation — see `docs/NETWORK_API_INVESTIGATION.md §3` for the full investigation.

### `GET /network/eras` — Conway `end` null

The Protocol Version 11 intra-era hardfork on preprod has not yet fired at the time of testing. The preprod chain was at epoch 284, slot ~121,280,785; the hardfork is scheduled at slot 121,478,400 (epoch 285 start, ~2.3 days away). yaci-store reports `null` because no successor era row exists in the `era` table yet — `EraService.getEras()` only returns eras that have been observed on-chain.

Blockfrost returns `{"time": 123120000, "slot": 121478400, "epoch": 285}` because its backend (cardano-db-sync) reads the ratified `HARD_FORK_INITIATION_ACTION` governance action and pre-computes the scheduled boundary before the hardfork fires.

This is **not a sync lag** — yaci-store was verified to be fully synced to the current chain tip (local max slot 121,280,706 vs chain tip 121,280,542). Once the hardfork fires, the `era` table will receive a new row and Conway's `end` will be populated automatically. No code change needed.

---


extensions/blockfrost/network/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/network/service/BFNetworkService.java
```diff
@@ -0,0 +1,206 @@
...
+                .maxKesEvolutions(MAX_KES_EVOLUTIONS)
+                .securityParam(genesisConfig.getSecurityParam())
+                .build();
+    }
+
+    // ── / (root) ─────────────────────────────────────────────────────────────
+
+    public BFRootDto getRoot() {
+        return BFRootDto.builder()
+                .url("https://cardano-mainnet.blockfrost.io/api")
```
Sotatek-HuyLe3a: do it always return mainnet url?

satran004: I don't think this should point to Blockfrost. May be we need to keep bf_hostname as a config and add path accordingly.

extensions/blockfrost/network/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/network/service/BFNetworkService.java
```diff
@@ -0,0 +1,206 @@
...
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class BFNetworkService {
+
+    private static final String BF_API_VERSION = "0.1.30";
+
+    // Standard Cardano protocol constants (identical across mainnet, preprod, preview)
+    private static final long SLOTS_PER_KES_PERIOD = 129600L;
```
satran004: These values should come from Genesis file (Genesis class) instead of hardcoded value.

adamcazes:
Two small findings about the root endpoint from the same mainnet validation run as the #1079 era check, recorded here since this branch owns `getRoot()`:

1. The official blockfrost-js SDK requests the API root with a trailing slash. Spring Boot 3 defaults `trailing-slash-match` to false, so `GET {prefix}/` returns 404 while `GET {prefix}` returns 200. Result: the SDK's `root()` method still fails against the module. Mapping both variants (or an explicit `"/"` path on the controller) would let it pass.

2. `getRoot()` returns `hostname + apiPrefix`. With `blockfrost.hostname` unset the `url` field comes back relative (`"/api/v1/blockfrost"`), while hosted Blockfrost returns an absolute URL. Worth a config-doc note that `blockfrost.hostname` should be set for an absolute `url`.

Happy to push either change to this branch as a follow-up commit if you want them handled here rather than separately.

Sotatek-HuyLe3a:
Thanks @adamcazes, both root endpoint findings are handled in this branch now.

  - Fixed the trailing slash case in `BFNetworkController`: `GET {prefix}` and `GET {prefix}/` now both map to `getRoot()`. Commit: `80ed5edd5`
  - Added docs for `blockfrost.hostname` in the Blockfrost-compatible configuration page. The docs now clarify that this is the yaci-store Blockfrost-compatible root endpoint, and that leaving `blockfrost.hostname` empty returns
  a relative `url`, while setting it to the yaci-store public origin returns an absolute `url`. Commit: `ee7084276`

satran004 (approved):

PR #870, opened 2026-03-27
feat: assets-ext — token metadata (CIP-26 + CIP-68 [FT, NFT] + CIP-113)

matiwinnetou:
## Summary

Adds a new extension module (`extensions/assets-ext`) that indexes and serves Cardano token metadata, merging off-chain (CIP-26) and on-chain (CIP-68) sources transparently. Replaces the need for a standalone [cf-token-metadata-registry](https://github.com/cardano-foundation/cf-token-metadata-registry) deployment.

### Supported Standards

| Standard | Source | Indexing | API |
|----------|--------|----------|-----|
| **CIP-26** | Off-chain — GitHub [cardano-token-registry](https://github.com/cardano-foundation/cardano-token-registry) | Active | Exposed |
| **CIP-68 FT** (label 333) | On-chain — reference NFT inline datums | Active | Exposed |
| **CIP-68 NFT / RFT** (labels 222 / 444) | On-chain — reference NFT inline datums | Active (rows persisted into `cip68_metadata` alongside FTs) | **Not yet exposed** — schema and indexing in place; no public REST endpoints are wired up for these labels in this PR |
| **CIP-113** | On-chain — programmable token registry | Active | Exposed when enabled; **disabled by default** until CIP-113 is officially live on mainnet (set `store.assets.ext.cip113.enabled=true` to opt in on preview/preprod) |

### REST API

All endpoints use configurable `${apiPrefix}` (default `/api/v1`) and are rooted at `${apiPrefix}/tokens`.

**Merged metadata (cf-token-metadata-registry V2 API compatible)** — queries by subject return CIP-26+CIP-68 merged transparently with configurable priority. Response includes `type` field (`NATIVE` or `PROGRAMMABLE`) and optional `extensions.cip113` when CIP-113 is enabled:

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/tokens/subject/{subject}` | Merged metadata with CIP priority, property filtering, `show_cips_details` |
| `POST` | `/tokens/subject/query` | Batch query (max 100 subjects) with pre-fetched data (3 bulk queries) |

These endpoints follow the same contract as [cf-token-metadata-registry's V2 API](https://github.com/cardano-foundation/cf-token-metadata-registry), including:
- `query_priority` parameter (`CIP_68,CIP_26` default — on-chain preferred)
- `show_cips_details` flag for raw per-standard metadata
- `property` filtering with required `name` + `description`
- `type` field: `NATIVE` vs `PROGRAMMABLE`
- `extensions.cip113` block with `transfer_logic_script`, `third_party_transfer_logic_script` (each with a `*_type` companion `VKEY`/`SCRIPT`), `global_state_policy_id` (all nullable)

**CIP-specific (direct access — symmetric interfaces):**

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/tokens/cip26/{subject}` | CIP-26 metadata by subject |
| `GET` | `/tokens/cip26/{policyId}/{assetName}` | CIP-26 metadata by policy ID + asset name |
| `GET` | `/tokens/cip68/ft/{subject}` | CIP-68 FT metadata by subject (prefix resolved automatically) |
| `GET` | `/tokens/cip68/ft/{policyId}/{rawAssetName}` | CIP-68 FT metadata by policy ID + raw asset name (without CIP-68 label prefix) |

Note: there are no CIP-68 NFT (222) or RFT (444) endpoints in this PR. Rows are persisted (see below) so a follow-up can expose them without re-indexing.

### Known divergences from cf-token-metadata-registry V2

The V2-shaped endpoints are byte-equivalent to CF for almost every field. Two cases differ deliberately:

1. **`logo.value` encoding.** [CIP-26 spec](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0026/README.md) explicitly defines logos as `"encoding": "base64"`. The upstream registry stores base64; yaci-store passes it through unchanged. The reference CF V2 API instead decodes base64 and re-encodes as hex, which is non-spec. Consumers migrating from CF need to flip image rendering from `data:image/png;hex,...` to the standard `data:image/png;base64,...`.
2. **Tokens whose registry has multiple files claiming the same on-chain subject.** yaci-store deterministically picks the file whose filename equals the subject (the registry's documented naming convention) and skips the rest with a `WARN` log; CF picks non-deterministically based on `File.listFiles()` iteration order. On the testnet registry today this affects 3 subjects (notably `baa836fef0…` / BCoin and `7755534454` / wUSDT). On mainnet this affects zero tokens — every mainnet file's name matches its subject.

Everything else — schema, query parameters, source attribution, `extensions.cip113`, batch endpoint shape — matches CF byte-for-byte. Mainnet verification at the current chain tip: **7,932 / 7,932 CIP-26 subjects** and **7,647 / 7,647 CIP-68 FT subjects** byte-identical against the CF V2 API.

### Programmatic API (StorageReaders)

For projects using the starter as a library dependency without REST:

| Interface | Key methods |
|-----------|-------------|
| `AssetsReader` (facade) | `getSubject()` (3 overloads), `getSubjects()` (batch), `getCip26Metadata()`, `getCip26Logo()`, `getCip68Metadata()`, `getCip113RegistryNode()`, `getCip113RegistryNodes()` (batch), `isProgrammableToken()` |
| `Cip26StorageReader` | `findBySubject()`, `findBySubjects()`, `findLogoBySubject()`, `findLogosBySubjects()` |
| `Cip68StorageReader` | `findBySubject()`, `findBySubject(subject, properties)`, `findByPolicyIdAndAssetName()`, `findBySubjects()` (batch). FT-facing today; NFT/RFT rows are in the table but not surfaced. |
| `Cip113StorageReader` | `findByPolicyId()`, `findByPolicyIds()` (batch), `isProgrammableToken()` |

### Database Schema

| Table | PK | Description |
|-------|----|-------------|
| `cip26_metadata` | `(subject)` | CIP-26 fungible-token metadata, logo (`TEXT`/`LONGTEXT` base64), and raw `Mapping` JSON (`properties`) — one row per CIP-26 subject. Merges what were previously three tables (`ft_offchain_metadata` + `ft_offchain_logo` + `off_chain_sync_state`). |
| `cip26_sync_state` | `(id)` | GitHub-registry sync progress tracking (last commit hash + `last_synced_at`). |
| `cip68_metadata` | `(policy_id, asset_name, slot)` | **All three CIP-68 labels** (222 NFT / 333 FT / 444 RFT) in a unified table. Composite PK preserves the full history of reference-NFT updates. Per-row `label` is set by `Cip68Processor` based on the co-minted user-token prefix observed in the same transaction's outputs. |
| `cip113_registry_node` | `(key, slot, tx_hash)` | CIP-113 programmable-token registry. `key` is the linked-list sort key — equal to the programmable token's 28-byte policy ID (56 hex chars) for real rows, sentinel marker for head/tail rows. |

JPA entity class names mirror the table names: `Cip26Metadata`, `Cip26SyncState`, `Cip68Metadata`, `Cip113RegistryNode`. Service-layer class names match the CIP they implement: `Cip26MetadataService`, `Cip26MetadataValidator`, `Cip26MetadataSyncService`, `Cip26MetadataSyncCronJob`, `Cip68TokenService`, `Cip113RegistryService`, `TokenQueryService` (the merged-view aggregator).

All tables include `last_synced_at` for monitoring. Flyway migration version `V0_1600_1` (renumbered from `V0_1500_1` to avoid collision with `analytics-store`). The three dialect-specific `V0_1600_1__init.sql` files (h2 / mysql / postgresql) hold just the DDL; per-column rationale, CIP references, and dialect notes live in [`db/store/README.md`](extensions/assets-ext/src/main/resources/db/store/README.md).

**`label` is per-row, not per-asset (read-path implication).** Because the classifier tags each reference-NFT update with whichever user-token was co-minted that transaction, a single reference NFT can have rows under multiple labels across its history (e.g. an FT renamed during an update co-minted with a 222 NFT will have its newest row stored under label=222). Read-path queries therefore **do not filter by `label`** — `Cip68MetadataRepository.findFirstByPolicyIdAndAssetNameOrderBySlotDesc` and `findLatestByConcatenatedKeys(keys)` both return the latest row by slot regardless of label.

**Index strategy:** Default DDL includes rollback indexes (`slot`). CIP-113 PK `(key, slot, tx_hash)` supports backward scan for latest-by-key queries without a separate index. Query indexes (policy, ticker, label) are in `optional-indexes.sql` — apply per use case.

### CIP-26 sync hardening

The sync path was hardened during QA against the testnet registry (where ~5,008 of 5,526 files are filename≠subject duplicates) to be deterministic and self-recovering:

- **`InsertOutcome` enum** (`INSERTED` / `PERMANENTLY_SKIPPED` / `TRANSIENTLY_FAILED`). Save failures are now classified — `NonTransientDataAccessException` (constraint violation, value-too-long) → permanently skipped, log loud, advance cursor; `TransientDataAccessException` (lock timeout, lost connection) → block cursor, retry next sync; unknown exceptions → conservative default = transient. This prevents both the silent-data-drop class of bugs and the cursor-stuck-forever class.
- **Filename-vs-inner-subject filter.** A registry file is indexed only if `filename == json["subject"]`. Filenames are unique per directory, so each subject deterministically maps to one canonical file. Mismatched files are logged as `WARN Skipping 'X.json': filename does not match inner subject 'Y'` and counted in the completion log line. This eliminates `File.listFiles()`-order non-determinism that surfaced as different per-token content across deployments.
- **Per-network defaults resolved from `protocol-magic`.** The starter's `AssetsExtProperties.Cip26.gitOrganization`/`gitProjectName`/`gitMappingsFolder` default to `null`; `Cip26NetworkDefaults` then maps protocol-magic to the right upstream (mainnet → cardano-foundation, preprod → input-output-hk). An earlier hardcoded-mainnet default in the starter caused preprod to silently index mainnet data — locked down by `AssetsExtPropertiesTest`.
- **Existing-clone URL mismatch.** `GitService` re-clones if the on-disk folder's `remote.origin.url` doesn't match the configured one. Operators switching networks no longer need to remember `force-clone=true`.

### Read-Only Mode

The extension works in both write and read-only (`store.read-only-mode=true`) modes:

- **Write mode**: all processors, cron job, and health indicators active
- **Read-only mode**: REST API and StorageReaders serve queries from existing data. Write-path components are gated:
  - `Cip26MetadataSyncCronJob`: `@ReadOnly(false)` — prevents `@Scheduled` timer from firing
  - `OffchainSyncHealthIndicator`: `@ConditionalOnBean(Cip26MetadataSyncCronJob.class)` — silently skipped
  - Node-connection / readiness indicators: `@ConditionalOnBean(HealthService.class)` — silently skipped

### Rollback Support

Both CIP-68 and CIP-113 processors handle `RollbackEvent`:
- Delete entries with `slot > rollbackSlot` via a single bulk `@Modifying @Query` DELETE (no per-row round-trip)
- Composite PK with slot allows historical tracking — `ORDER BY slot DESC` picks correct pre-rollback entry

### Performance

- **Batch endpoint**: `prefetchBatch()` issues 3 bulk queries upfront (CIP-26 metadata, CIP-26 logos, CIP-113), then O(1) map lookups per subject. 50 subjects = ~53 queries (down from ~200)
- **CIP-68 batch query**: `ROW_NUMBER() OVER (PARTITION BY policy_id, asset_name ORDER BY slot DESC)` — single-pass window function returning the latest reference-NFT row per `(policy, asset_name)` regardless of label.
- **CIP-113 batch query**: same window-function pattern for `(policy_id, slot DESC)`.
- **CIP-26 sync**: batch JGit history resolution — single git log walk for all files (O(commits × tree-diff)) instead of per-file `git log --path` (O(files × commits))
- **Processor writes**: both CIP-68 and CIP-113 use `saveAll()` for Hibernate JDBC batch inserts
- **Logo skip**: logo query skipped when not in requested properties

### CIP-113 Alignment with cf-token-metadata-registry

- All three script fields (`transfer_logic_script`, `third_party_transfer_logic_script`, `global_state_policy_id`) are nullable — parser only requires `key` and `next` fields
- **Aiken `Credential` variant preserved.** Both transfer-logic-script columns now have a `*_type` companion (`VARCHAR(8)`, values `VKEY` or `SCRIPT`) so downstream verifiers can distinguish vkey-signature checks from script-witness execution. Without this column the variant is unrecoverable post-persistence — adding it later would require ALTER TABLE plus reparsing every datum row. The DTO surfaces it as `transfer_logic_script_type` / `third_party_transfer_logic_script_type`. Hibernate `@Enumerated(EnumType.STRING)` is used so the column is queryable from raw SQL without Java enum awareness; the enum is registered in `reflect-config.json` for the GraalVM native build.
- `key` column stores the sort-key field from the datum (first field in the linked list node). For real registrations it equals the 56-hex policy ID of the registered programmable token; for the two sentinel rows per registry (head = empty string, tail = conventionally 32 bytes of `0xFF`) it is a linked-list marker that is never a policy ID
- PK `(key, slot, tx_hash)` supports historical tracking
- Empty strings normalized to null in DTO conversion

### CIP-68 NFT / RFT indexing (no public API yet)

The `cip68_metadata` table holds **all three label types** (222 NFT / 333 FT / 444 RFT) — they share the same table, schema, and processor. `Cip68Processor` classifies every reference-NFT update at index time using cross-output detection: it inspects the user-token prefixes co-minted in the same transaction's outputs and tags the row's `label` column accordingly. NFT-specific columns (`image`, `media_type`) and FT-specific columns (`decimals`, `ticker`, `logo`) are nullable so the same row schema accommodates all three.

What is shipped in this PR:
- Schema, processor, parser, and storage all handle 222/333/444 uniformly.
- The `properties` JSON column (PG `JSONB` / MySQL `JSON` / H2 `TEXT`) stores collection-specific structures like `files[]` and arbitrary trait/attribute maps, so NFT data is round-trippable.

What is **not** shipped:
- No REST endpoints under `/tokens/cip68/nft/...` or `/tokens/cip68/rft/...`.
- `Cip68StorageReader` only returns `FungibleTokenMetadata`; no NFT/RFT-shaped DTO or service method.

A follow-up PR can add the public surface without re-indexing — the data is already on disk.

### Configuration

```yaml
store:
  assets:
    ext:
      enabled: true               # master switch (default: false)
      cip26:
        enabled: false            # default: false — gates the GitHub sync cron job (breaking change vs earlier draft of this PR; previously defaulted to true)
        sync-interval-minutes: 60 # default: 60 — @Scheduled fixedDelay
        force-clone: false        # default: false — wipe + re-clone git registry on startup
      cip68:
        enabled: true             # default: true — gates the Cip68Processor / Cip68RollbackProcessor event listeners
      cip113:
        enabled: false            # default: false — until CIP-113 is officially live on mainnet
      query:
        priority: "CIP_68,CIP_26"
```

CIP-68 uses `@ConditionalOnProperty(matchIfMissing = true)`. CIP-26 and CIP-113 use `matchIfMissing = false` — the sync cron and CIP-113 processor beans aren't registered unless the operator explicitly opts in. Storage readers and the REST API stay active so any already-persisted data keeps serving.

Network-specific config (git URLs, CIP-113 policy IDs) is auto-detected from `store.cardano.protocol-magic`. Per-network policy ID files in `resources/cip113/cip113-{network}.properties` carry a disclaimer that the values change once CIP-113 hits mainnet.

When the in-process CIP-26 sync is off, `SyncStatusEnum.SYNC_DISABLED` (renamed from the misleading `SYNC_IN_EXTRA_JOB`) is reported and `OffchainSyncHealthIndicator` maps it to `Health.UP` — operators who turn the cron off are assumed to be keeping the registry fresh externally; returning `DOWN` would falsely page on every probe.

### Architecture

- **Module**: `extensions/assets-ext` (artifact: `yaci-store-assets-ext`)
- **Starter**: `starters/assets-ext-spring-boot-starter`
- **Properties**: `store.assets.ext.*` (avoids collision with existing `store.assets` native asset store)
- **No `@Value` in extension module** — all config flows through `AssetsExtStoreProperties` POJO, mapped by starter's `AssetsExtProperties` AutoConfiguration (follows yaci-store convention)
- **Services use StorageReader interfaces**, not repositories directly (replaceable via `@ConditionalOnMissingBean`)
- **CIP-113 activation**: storage reader and configuration always registered; processor/rollback-processor beans gated by `store.assets.ext.cip113.enabled` (off by default). Effective `Cip113Configuration.isEnabled()` reflects whether the per-network registry NFT policy-id set is non-empty (used by the storage reader to short-circuit empty queries).

## Test plan

- [x] Unit tests for CIP-113 parser, processor, service, storage reader (nullable fields + VKEY/SCRIPT credential type discriminator)
- [x] Unit tests for `Cip68TokenService` (prefix conversion, reference NFT detection, latest-row-regardless-of-label contract)
- [x] Unit tests for `Cip26MetadataValidator` (CIP-26 field and logo validation)
- [x] Unit tests for CIP-68 parser, processor, asset type
- [x] Unit tests for `TokenQueryService` (single + batch)
- [x] Unit tests for `Cip26MetadataService` exception classification (validation rejected, non-transient DB error, transient DB error, unknown exception, happy path)
- [x] Unit tests for `Cip26MetadataSyncService` (sync states, incremental, full, transient blocks cursor, permanent advances cursor, **filename-mismatch filter**, `SYNC_DISABLED` when cip26 sync is off)
- [x] Unit tests for `Cip26NetworkDefaults` (mainnet / preprod / preview / sanchonet / devkit defaults + user overrides)
- [x] Unit tests for `GitService` (root-commit regression, **remote-URL mismatch detection**)
- [x] Regression-lock test for starter `AssetsExtProperties` null-defaults convention
- [x] Starter compiles
- [x] SonarCloud issues resolved
- [x] **Mainnet verification (full sweep at chain tip):** 7,932 / 7,932 CIP-26 subjects and 7,647 / 7,647 CIP-68 FT subjects byte-identical against the CF V2 API. Verified via `test-scripts/verify_assets_ext_cip26.py` (preprod earlier: 511/512).
- [x] Filename-mismatch filter verified on preprod: 5,008 of 5,526 mismatched files correctly skipped, 515 canonical entries indexed, deterministic across reruns
- [x] CIP-68 indexing end-to-end on mainnet (27,544 rows across labels 222/333/444 in `cip68_metadata` at sync completion)


config/application.properties
```diff
@@ -350,9 +350,16 @@ blockfrost.apiPrefix=/api/v1/blockfrost
...
 #store.extensions.blockfrost.asset.enabled=false
-#store.extensions.blockfrost.blocks.enabled=false
 server.forward-headers-strategy=framework
 
+#######################################################################################
+# Asset Store Extension (CIP-26 + CIP-68 + CIP-113 Token Metadata)
+# CIP-26 git registry and CIP-113 policy IDs are auto-detected from
+# store.cardano.protocol-magic. CIP-68 is on-chain and needs no config.
+#######################################################################################
+#store.extensions.asset-store.enabled=false
```
satran004: Update the property name based on final module name or just use `assets`.  Fyi, the core store modules doesn't have `store` suffix.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/processor/Cip113EventListener.java
```diff
@@ -0,0 +1,66 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class Cip113EventListener {
```
satran004: The naming convention used in yaci-store for EventListener is `xxxxxProcessor`. So this can be changed to `Cip113EventProcessor`.

extensions/asset-store/src/main/resources/db/store/h2/V0_1500_1__init.sql
```diff
@@ -0,0 +1,60 @@
...
+    updated     TIMESTAMP,
+    updated_by  VARCHAR(255),
+    properties  TEXT
+);
+
+-- CIP-26 logos
+CREATE TABLE token_logo (
+    subject VARCHAR(255) PRIMARY KEY,
+    logo    TEXT,
+    CONSTRAINT fk_token_logo_metadata_subject FOREIGN KEY(subject) REFERENCES token_metadata(subject)
```
satran004: If possible, let's remove FK and maintain data consistency through application layer/rollback strategy.

extensions/asset-store/src/main/resources/db/store/h2/V0_1500_1__init.sql
```diff
@@ -0,0 +1,60 @@
...
+    version     BIGINT NOT NULL,
+    datum       TEXT NOT NULL,
+    PRIMARY KEY (policy_id, asset_name, slot)
+);
+
+CREATE INDEX idx_metadata_reference_nft_slot ON metadata_reference_nft(slot);
+
+-- CIP-113 programmable token registry nodes
+CREATE TABLE cip113_registry_node (
+    policy_id                         TEXT   NOT NULL,
```
satran004: Same as before

extensions/asset-store/src/main/resources/db/store/h2/V0_1500_1__init.sql
```diff
@@ -0,0 +1,60 @@
...
+    slot                              BIGINT NOT NULL,
+    tx_hash                           TEXT   NOT NULL,
+    transfer_logic_script             TEXT,
```
satran004: If this is script_hash, this could be `varchar(56)`

Fyi: script store DDL https://github.com/bloxbean/yaci-store/blob/main/stores/script/src/main/resources/db/store/postgresql/V0_400_1__init.sql

extensions/asset-store/src/main/resources/db/store/optional-indexes.sql
```diff
@@ -0,0 +1,24 @@
+-- Optional indexes for the asset-store extension.
+-- These are NOT applied automatically. Apply them manually if your queries need them.
+-- See also: store.auto-index-management in yaci-store for the core index management pattern.
```
satran004: Auto index management is no longer supported.

1. Yaci Store users should use the admin CLI tool to apply indexes after sync.

2. External applications like Rosetta can implement their own process to apply indexes after sync by monitoring the sync process.

We can keep this file as-is for now. Later, we will integrate it into the admin CLI.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/api/controller/Cip68MetadataController.java
```diff
@@ -0,0 +1,42 @@
...
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping("${store.extensions.asset-store.api-prefix:/api/v1}")
```
satran004: We already have `${apiPrefix}` config which is used by other modules. I think we can re-use it instead introducing new one.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/AssetStoreExtConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+        havingValue = "true",
+        matchIfMissing = false
+)
+@ComponentScan(basePackages = {"com.bloxbean.cardano.yaci.store.extensions.assetstore"})
+@EnableJpaRepositories(basePackages = {"com.bloxbean.cardano.yaci.store.extensions.assetstore"})
+@EntityScan(basePackages = {"com.bloxbean.cardano.yaci.store.extensions.assetstore"})
+@EnableTransactionManagement
+@EnableScheduling
+public class AssetStoreExtConfiguration {
+    public static final String STORE_ASSET_STORE_EXT_ENABLED = "store.extensions.asset-store.enabled";
```
satran004: Please check if it's used anywhere, otherwise please remove.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/processor/Cip113EventListener.java
matiwinnetou: Fixed — renamed to `Cip113Processor`.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/entity/Cip113RegistryNode.java
```diff
@@ -0,0 +1,50 @@
...
+@AllArgsConstructor
+@Builder
+public class Cip113RegistryNode {
+
+    @Id
+    @Column(name = "policy_id")
+    private String policyId;
+
+    @Id
+    private Long slot;
```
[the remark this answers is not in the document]
matiwinnetou: Added slot to PK — now `(policy_id, slot, tx_hash)`. This enables backward scan for latest-by-policy queries and matches cf-token-metadata-registry.

extensions/asset-store/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/api/controller/Cip113RegistryController.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+import java.util.Map;
+
+@RestController
+@RequestMapping("${store.extensions.asset-store.api-prefix:/api/v1}")
+@ConditionalOnExpression("${store.extensions.asset-store.enabled:false}")
```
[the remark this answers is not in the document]
matiwinnetou: Fixed — `Cip113RegistryController` removed. CIP-113 data surfaces through the merged subject response as an extension. No separate controller or condition needed.

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip68/model/AssetType.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Record/Utility class to group in one place Asset unit, policy and name manipulation and comparison
+ *
+ * @param policyId  the 56-character hex policy ID
+ * @param assetName the hex-encoded asset name (may be empty for policy-only assets)
+ */
+@Slf4j
+public record AssetType(String policyId, String assetName) {
```
matiwinnetou, line 12: Do we have a similar utility class in Yaci-Store itself?

components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/StoreConfiguration.java
```diff
@@ -80,4 +82,16 @@ public LocalClientProviderManager localClientProviderManager(Environment env,
...
+
+    /**
+     * Shared {@link Clock} for time-stamping persisted rows. Returns
+     * {@link Clock#systemDefaultZone()} — bytecode-equivalent to {@code LocalDateTime.now()},
+     * which preserves existing behavior across modules. Override (e.g. {@code Clock.fixed(...)})
+     * in tests for deterministic time assertions.
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    public Clock systemClock() {
```
matiwinnetou: @satran004 I hope you agree with this, I only am adding this here and we can then use Clock and refactor other modules to inject Clock instance.

config/application.properties
```diff
@@ -353,6 +353,17 @@ blockfrost.apiPrefix=/api/v1/blockfrost
 #store.extensions.blockfrost.blocks.enabled=false
 server.forward-headers-strategy=framework
 
+#######################################################################################
+# Assets Extension (CIP-26 + CIP-68 + CIP-113 Token Metadata)
+# CIP-26 git registry auto-detected from store.cardano.protocol-magic.
+# CIP-68 is on-chain and needs no config.
+# CIP-113 registry NFT policy IDs auto-detected from protocol-magic.
+#######################################################################################
+store.assets.ext.enabled=true
```
Sotatek-HuyLe3a: Could we keep `store.assets.ext.enabled` and `store.assets.ext.cip26.enabled` disabled in the config?  I think we can  keep these values commented or set to false

extensions/assets-ext/src/main/resources/db/store/postgresql/V0_1600_1__init.sql
```diff
@@ -0,0 +1,65 @@
...
+    url            VARCHAR(250),
+    decimals       BIGINT,
+    logo           TEXT,
+    image          TEXT,
+    media_type     VARCHAR(255),
+    version        BIGINT       NOT NULL,
+    datum          TEXT         NOT NULL,
+    properties     JSONB,
+    last_synced_at TIMESTAMP,
+    PRIMARY KEY (policy_id, asset_name, slot)
```
Sotatek-HuyLe3a: is it possible for the same CIP-68 reference NFT to be updated more than once in the same slot? The current PK is `(policy_id, asset_name, slot)`, so if that case can happen, we may need transaction/output identity in the key to avoid duplicate inserts and preserve full history.

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/storage/impl/repository/Cip113RegistryNodeRepository.java
```diff
@@ -0,0 +1,62 @@
...
+     * reserved-word {@code key} identifier does not need dialect-specific quoting in the SQL.
+     *
+     * <p>The correlated {@code MAX(slot)} subquery is efficient given the
+     * {@code (key, slot)} portion of the primary key; PostgreSQL / H2 / MySQL all plan this
+     * as an index-backed per-key lookup.
+     */
+    @Query("SELECT e FROM Cip113RegistryNode e " +
+            "WHERE e.key IN :keys " +
+            "AND e.slot = (SELECT MAX(e2.slot) FROM Cip113RegistryNode e2 WHERE e2.key = e.key)")
+    List<Cip113RegistryNode> findLatestByKeys(@Param("keys") Collection<String> keys);
```
Sotatek-HuyLe3a, line 55: `findLatestByKeys()` selects rows where `slot = MAX(slot)`, while the table key allows multiple rows for the same key in the same slot via `(key, slot, tx_hash)`. So, if two records have the same (key, slot), is this still safe in our context?

config/application.properties
matiwinnetou: Good point, thanks — fixed, though I went a slightly different route and I'd like your read on it.

The behaviour follows the yaci-store convention that activation is driven by whether the module's starter is on the classpath, the same as the core stores. `BlocksStoreConfiguration`, `UtxoStoreConfiguration`, `TransactionStoreConfiguration`, `MetadataStoreConfiguration` and `AssetsStoreConfiguration` are all `@ConditionalOnProperty(..., matchIfMissing = true)` — so for an embedded integrator, adding the starter dependency is itself the opt-in, and they don't have to set anything. The aggregates that have heavier/optional side effects (e.g. `governance-aggr`) use `matchIfMissing = false` instead and must be enabled explicitly.

So I aligned assets-ext to that split rather than gating the whole extension behind an explicit flag:

- **Master gate** (`AssetsExtConfiguration`) → `matchIfMissing = true`: the extension auto-enables on the classpath, matching the core stores. The work that runs by default is **CIP-68 only**, which is pure on-chain datum parsing — cheap, no external calls.
- **CIP-26** → stays `matchIfMissing = false` (off by default). It runs a cron that clones/pulls the off-chain GitHub registry, so I don't want merely adding the dependency to start network sync — it's an explicit opt-in.
- **CIP-113** → stays `matchIfMissing = false` (off by default) since it's not yet live on mainnet.

For the **standalone build**, which bundles every starter, I set all four `store.assets.ext.*` flags to `false` in both `config/application.properties` and `application-all.properties` — so the standalone distribution ships with the extension off and an operator flips one flag to opt in. That's exactly the disabled-by-default state you asked for; the auto-enable only applies to embedded integrators who deliberately pull in the `assets-ext` starter.

Net effect: nothing with external side effects (CIP-26 GitHub sync, CIP-113) ever starts implicitly, and the standalone app is off out of the box. Let me know if you'd rather have the master flag be explicit-opt-in too (i.e. `matchIfMissing = false` like `governance-aggr`) — happy to switch if you think the parity with core stores is the wrong call here.

satran004: I think, in this case we should align it with blockfrost support which is also an extension. In case of blockfrost, the master flag is disabled by default and all other sub flags are enabled by default. So user just needs to enable the master flag to get the default behavior.
So, we can keep all asset-ext properties commented in application.properties.

extensions/assets-ext/src/main/resources/db/store/postgresql/V0_1600_1__init.sql
satran004: May be tx_index will help here. But this scenario is possible.

matiwinnetou: Good question, and yes that case is possible — but it's by design that the current PK `(policy_id, asset_name, slot)` is what we want here.

**It can happen, but it's extremely rare.** A CIP-68 reference NFT (label 100) is a quantity-1 token living in a single UTxO, so to update its metadata twice within the same slot you'd need two transactions in the *same block* where the second spends the output produced by the first — i.e. intra-block transaction chaining. Cardano does allow that, so it's not impossible, but in practice it's a very uncommon pattern for reference-NFT metadata updates.

**When it does happen, the behaviour is last-writer-wins, and that's exactly what we want.** `Cip68Processor` walks the block's outputs in order and `saveAll`s them. Two updates at the same slot collapse to the same primary key, so the second one merges over the first — we end up with a single row holding the *latest* in-block state. So there is no duplicate-insert and no exception; the PK does the dedup for us, and the chronological processing order guarantees the surviving row is the most recent update in that slot. The readers (`findFirstByPolicyIdAndAssetNameOrderBySlotDesc` and the `ROW_NUMBER() ... ORDER BY slot DESC` batch query) then return that one latest row.

It would actually be a *problem* if we did **not** have last-writer-wins here — e.g. if the key allowed multiple same-slot rows and a reader picked an arbitrary one, you could surface stale metadata. Keeping `tx_hash`/output identity out of the key is what gives us the clean "one deterministic latest row per asset per slot" guarantee.

**On "preserve full history":** that's intentionally out of scope. The goal of the asset store is to serve the *current* resolved metadata for an asset, not to be an audit trail of every intermediate datum. We keep per-slot history (so rollbacks and point-in-time-by-slot work), but we deliberately don't retain multiple updates *within* a single slot — that intermediate state has no consumer for a current-state store, and collapsing it is the desired outcome rather than a loss.

(Worth noting this is the deliberate counterpart to the CIP-113 keying you flagged separately: there `tx_hash` *is* in the key, which is why that reader needs an explicit tie-break. CIP-68 sidesteps that whole class of issue by keying on slot only.)

satran004: I agree that tx chaining might be rare in this case, but it is quite common on mainnet for other usecases. Actually, we were forced to keep tx_index for all other scenarios like stake_registration etc to correctly calculate stake amount and adapot. In all other places, we keep history and don’t do inplace update. I am just wondering if we should break that guideline here. May be a simple addition of tx_index column keep it compatible with overall design. Tx index is already part of transaction object in event.

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/storage/impl/repository/Cip113RegistryNodeRepository.java
matiwinnetou: Great catch — this was a real latent bug, fixed in 5bf2ed3b9.

You're right: with the PK `(key, slot, tx_hash)`, two updates to the same registry node in the same slot produce two rows at `MAX(slot)`, so `findLatestByKeys()` returned both, and `Cip113StorageReaderImpl.findByPolicyIds()` would then throw `IllegalStateException` from `Collectors.toMap` on the duplicate key. Extremely rare (it needs intra-block transaction chaining — a second tx in the same block spending the first's registry-node output) but a hard crash when it happens.

**Fix: drop `tx_hash` from the primary key, so the PK is `(key, slot)`.** `tx_hash` stays as a provenance column, just not part of the key.

Why this approach rather than de-duping at read time:

- **It resolves the conflict at write time, deterministically and *correctly*.** Same-slot updates now collapse via `saveAll` merge in block/tx processing order, so the surviving row is the genuinely latest state (last-writer-wins). The alternative — keeping `tx_hash` in the key and adding a merge function to `toMap` — would have to pick a winner at *read* time, but by then there's no column that encodes in-block tx order (`slot` ties and `tx_hash` is a hash, not a sequence), so any tie-break would be deterministic-but-arbitrary and could surface the wrong state. Resolving at write time avoids that entirely.
- **It restores the "one latest row per key" guarantee the readers already assume.** With `(key, slot)` there can be at most one row per `(key, slot)`, so `findLatestByKeys`' `slot = MAX(slot)` resolves each key to exactly one row and `toMap` can never collide.
- **It's consistent with `cip68_metadata`**, which keys on `(policy_id, asset_name, slot)` for the same reason. The asset store is a current-state store, not a per-transaction audit trail, so collapsing same-slot updates is the intended behaviour rather than a loss.

No query/consumer/processor changes were needed — `tx_hash` remains an entity column with its getter, so existing tests are unaffected; the readers already resolved one row per key, which is now actually guaranteed. Schema is the init migration (`V0_1600_1__init.sql`, all three dialects) and the extension isn't released yet, so it's edited in place — no `ALTER` migration.

config/application.properties
matiwinnetou: Good call — adopted the blockfrost-extension pattern.

- **Master flag off by default**: `AssetsExtConfiguration` is now `@ConditionalOnProperty(name = "store.assets.ext.enabled", matchIfMissing = false)`, same as `BFAutoConfiguration`.
- **Sub-flags on by default**: the per-CIP gates default `matchIfMissing = true`, so enabling the single master flag gives the default behaviour without touching anything else.
- **Properties commented out** in `config/application.properties` (mirroring the `application-blockfrost.properties` style), and the standalone `application-all.properties` keeps the master flag explicitly off.

One deliberate deviation to flag: **CIP-113 stays off by default** (`matchIfMissing = false`). It isn't live on mainnet yet and we have an explicit team decision + a regression test (`cip113IsDisabledByDefault`) locking that, so enabling the master flag brings up **CIP-26 + CIP-68** by default and leaves CIP-113 as an explicit opt-in. CIP-26 is on by default per your guidance — the safeguard against a preprod node pulling mainnet metadata is the per-network git-registry resolution (`Cip26NetworkDefaults`, driven by `protocol-magic`), not a default-off flag.

extensions/assets-ext/src/main/resources/db/store/postgresql/V0_1600_1__init.sql
matiwinnetou: Agreed — kept the history and added `tx_index` rather than doing in-place updates. Done for both CIP-68 and CIP-113 (commits 9900c0eb0 and 77076fe12).

**Keying, aligned with the rest of yaci-store** (stake_registration / governance / pool / drep all key on `tx_hash`, with `tx_index` as a non-key ordering column):

- `cip68_metadata` PK → `(policy_id, asset_name, slot, tx_hash)`
- `cip113_registry_node` PK → `(key, slot, tx_hash)`
- `tx_index` is a non-key `NOT NULL` column on both. Latest state resolves by `ORDER BY slot DESC, tx_index DESC`, so same-slot intra-block updates are preserved as distinct rows and still disambiguate deterministically — no in-place overwrite.

**One core change worth your eyes** (in `9900c0eb0`, outside assets-ext): these processors consume `AddressUtxoEvent`, which didn't carry the tx index. I added a `txIndex` field to `AddressUtxo`, populated in `UtxoProcessor` / `ByronUtxoProcessor` by counting position in the block's tx list (the same canonical order `TransactionProcessor` uses for its `tx_index`). It's an event/domain-only field like `blockNumber`/`blockTime` — **not** persisted in `address_utxo`, no migration to the utxo store. Happy to adjust if you'd rather source it differently.

As a bonus this also resolves the duplicate-row issue @Sotatek-HuyLe3a flagged on CIP-113: `findLatestByKeys` now pins each key to `MAX(slot)` then `MAX(tx_index)`, returning exactly one row per key, so `Cip113StorageReaderImpl.findByPolicyIds` can't hit a `Collectors.toMap` collision — while keeping full history instead of collapsing to last-writer-wins.

matiwinnetou: Quick follow-up: after @satran004's input we changed direction here and made CIP-68 preserve full history rather than collapse same-slot updates to last-writer-wins.

`cip68_metadata` is now keyed `(policy_id, asset_name, slot, tx_hash)` with `tx_index` as a non-key ordering column (consistent with the rest of yaci-store — stake_registration, governance, etc. all key on `tx_hash`). So the same-slot intra-block update case you raised is now retained as distinct rows, and the latest state resolves deterministically via `ORDER BY slot DESC, tx_index DESC`. Done in 77076fe12.

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/storage/impl/repository/Cip113RegistryNodeRepository.java
matiwinnetou: Update — superseding my reply above. After @satran004's guidance to preserve full history (consistent with the rest of yaci-store) rather than collapse same-slot updates, I reversed the `(key, slot)` change.

Final design (in 77076fe12):

- `cip113_registry_node` PK → `(key, slot, tx_hash)` — `tx_hash` is back in the key, and `tx_index` (the producing tx's index within its block) is added as a non-key `NOT NULL` ordering column. So same-slot intra-block updates are kept as distinct rows instead of collapsing.
- `findLatestByKeys` now pins each key to `MAX(slot)` then `MAX(tx_index)`, returning exactly one row per key — so `Cip113StorageReaderImpl.findByPolicyIds` still can't hit the `Collectors.toMap` duplicate-key crash you flagged.

Net: the crash you found is still fixed, but now by deterministic latest-row resolution over preserved history, rather than by dropping `tx_hash` from the key. Same approach was applied to CIP-68.

extensions/assets-ext/src/main/resources/db/store/postgresql/V0_1600_1__init.sql
Sotatek-HuyLe3a: LGTM! Btw, could you rename the script file since we already have prefix V0_1600_1 in aggregates/epoch-nonce?

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/storage/impl/repository/Cip113RegistryNodeRepository.java
huylevt: LGTM!

satran004: @huylevt Are we good to merge this PR?

If so, we should **squash and merge** it into `main`, and then cherry-pick the commit(s) to the `release/2.0.x` branch. But, I think we'll also need to make some manual changes in `2.0.x` later, since it doesn't have the `extensions` module.

huylevt: yes, it's good to merge. After @matiwinnetou  rename the V0_1600_1 files  since they currently conflict with epoch-nonce module, we can merge this PR.

huylevt (approved):

extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip113/storage/impl/repository/Cip113RegistryNodeRepository.java
matiwinnetou: Done — renamed to `V0_1700_1__init.sql` across all three dialects (postgresql/h2/mysql) in bc8f1e3b9. `V0_1600` was already taken by `aggregates/epoch-nonce`; `V0_1700` is the next free block. Content unchanged, just the version prefix. Should be good to merge now. Thanks for the review!

Issue #880, opened 2026-04-01
[Governance] SPO Voting Auto-Passes When All Stake Abstains


Sotatek-HuyLe3a:
## Summary

`SPOVotingEvaluator` auto-passes a governance vote whenever all active SPO stake is counted as abstain, regardless of the configured threshold. The correct behavior is that when no non-abstaining stake participates, the acceptance ratio is effectively 0 — which is less than any non-zero threshold, so the vote should **fail**.

**Root Cause Explanation:**

```java
// SPOVotingEvaluator.java:33-34
if (totalStake.equals(BigInteger.ZERO) || totalAbstain.equals(totalStake)) {
    return VotingStatus.PASS_THRESHOLD;  // Auto-pass when all abstain
}
```

When all SPO stake abstains, the denominator `total - abstain` would be zero, and Yaci Store auto-passes.

**What the Haskell ledger does:**

When `total == abstain`, the ratio `yes / (total - abstain)` = `0/0`. In carrdano-ledger, this evaluates as 0. For any non-zero threshold, `0 >= threshold` is False — the vote **fails**.

PR #881, opened 2026-04-01
fix(governance-rules): SPO vote do not pass when all stake abstain and threshold is non-zero

Sotatek-HuyLe3a:
#880  
## Summary

  Fix incorrect auto-pass in `SPOVotingEvaluator` when all active SPO stake is counted as abstain.

  Previously, the evaluator returned `PASS_THRESHOLD` whenever `totalAbstain == totalStake` (or `totalStake == 0`), without checking the configured threshold. This caused governance actions to incorrectly pass when no
  non-abstaining stake participated.

  The correct behaviour: when the denominator `(totalStake - totalAbstain)` is zero, the acceptance ratio is effectively 0. A ratio of 0 is below any non-zero threshold, so the result should be `NOT_PASS_THRESHOLD`. The
   only legitimate auto-pass is when the threshold itself is 0 (no votes are required for this action type).


satran004:
@Sotatek-HuyLe3a So, in Yaci Devkit, to auto-pass, we need to make threshold to zero?

satran004 (approved):

PR #906, opened 2026-04-21
Fix DRep stake distribution: align PV9 stale reverse-index clearing with ledger

Sotatek-HuyLe3a:
## Summary

This PR fixes Conway DRep distribution mismatches during bootstrap/PV9 by aligning `yaci-store` snapshot logic with `cardano-ledger` behavior.

The main issue is the PV9 stale reverse-entry case:

- a stake address delegates to `old_drep`
- the same address later re-delegates to `new_drep`
- ledger can still keep a stale reverse entry under `old_drep`
- when `old_drep` unregisters, ledger may clear the address's forward delegation

If `yaci-store` only follows the latest visible `delegation_vote`, it can keep counting stake that ledger has already cleared.

## What Changed

- reworked `DRepDistService` to model PV9/bootstrap stale reverse clearing with `ss_pv9_cleared_addresses`
- moved the expensive post-bootstrap PV9 clear detection out of the snapshot hot path by adding a persistent `drep_pv9_stale_clear_event` cache
- added `drep_pv9_stale_clear_event_cache` as a marker table, including `pv9_source_max_slot` for rollback-safe cache reuse
- kept bootstrap/PV9 snapshots on the bounded source query while `pv9MaxEpoch` is still moving
- made post-bootstrap snapshots derive `ss_pv9_cleared_addresses` from cached clear events plus the dynamic `after_del` check
- removed the hardcoded DRep exclusion mechanism and replaced it with ledger-aligned query logic
- made DRep lifecycle matching cred-type aware by treating `(drep_hash, cred_type)` as the effective identity
- updated bootstrap-to-PV10 filtering so bootstrap-era delegations are kept only when they still match ledger rules
- excluded `ss_pv9_cleared_addresses` from both credential DRep distribution inserts and `ABSTAIN` / `NO_CONFIDENCE` inserts
- replaced null-fragile `NOT IN` exclusions with `NOT EXISTS`
- expanded inline comments and regression tests so the stale reverse logic is easier to follow and maintain

## Why This Fix Is Needed

- removes stake-count mismatches caused by stale reverse entries during bootstrap/PV9
- avoids relying on network-specific hardcoded exclusions
- keeps DRep stake snapshots closer to actual ledger semantics
- removes the repeated full historical PV9 clear-event reconstruction from post-bootstrap snapshot calculations

## Implementation Notes

- `ss_drep_status` now tracks DRep registration and unregistration using both `drep_hash` and `cred_type`
- `drep_pv9_stale_clear_event` stores immutable PV9 clear events, not a permanent list of excluded addresses
- `ss_pv9_cleared_addresses` remains snapshot-specific because an address can delegate again after being cleared and must then be counted again
- the post-bootstrap `ss_pv9_cleared_addresses` query reads the event cache and checks for later delegations with `after_del`
- rollback invalidation is guarded by `pv9_source_max_slot`; the cache is invalidated only when `rollbackTo.slot <= pv9_source_max_slot`
- no new index is added to `delegation_vote`
- the clear-event query guards against false positives across later UNREG/REG lifecycles, intermediate stake deregistration, `ABSTAIN` / `NO_CONFIDENCE` transitions, and fresh delegations after old DRep unregistration


aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/DRepPv9ClearEventCacheService.java
```diff
@@ -0,0 +1,366 @@
...
+                         AND after_del.cert_index > e.unreg_cert_index))
+            )
+            """;
+
+    @Transactional
+    public synchronized void ensureCacheReady(int pv9MaxEpoch) {
+        var params = new MapSqlParameterSource()
+                .addValue("pv9_max_epoch", pv9MaxEpoch);
+
+        Number markerCount = jdbcTemplate.queryForObject(CACHE_MARKER_COUNT_QUERY, params, Number.class);
```
satran004, line 281: @Sotatek-HuyLe3a If for a custom testnet, there is no delegation records during bootstrap phase, what would be the behavior? `rebuildCache` will be called during every epoch?

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/DRepDistService.java
```diff
@@ -710,7 +719,11 @@ and exists (
                   where
                     sd.address IS NULL
                     """ + excludeDelegationCondition + """
-                    AND rd.address NOT IN (SELECT address FROM ss_pv9_cleared_addresses)
+                    AND NOT EXISTS (
```
satran004, line 722: Will it cover if a bootstrap phase address has new drep delegation post bootstrap?

Sotatek-HuyLe3a: Yes, it should cover that case.
 `ss_pv9_cleared_addresses` is snapshot-specific, not a permanent exclusion list. When we build it from `drep_pv9_stale_clear_event`, we exclude a cleared address only if there is no later `delegation_vote` after the  PV9 stale-clear event.

aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/DRepPv9ClearEventCacheService.java
Sotatek-HuyLe3a: No, `rebuildCache` should not be called every epoch.

 Even if there are no PV9/bootstrap delegation records, `rebuildCache()` still inserts a marker row into `drep_pv9_stale_clear_event_cache` with `event_count = 0` and `pv9_source_max_slot = 0`

satran004 (approved):

PR #907, opened 2026-04-22
fix(governance-aggr): Sanchonet alignments

Sotatek-HuyLe3a:
## Summary

This PR fixes several Sanchonet governance mismatches found.

It updates the bundled Sanchonet Conway genesis/config, treats Sanchonet as a public network for bootstrap-related logic, resets committee carry-forward after `NO_CONFIDENCE`, and fixes committee hot-key registration handling at ratification boundaries.

## Changes

- Updated bundled Sanchonet network files:
  - `components/common/src/main/resources/store/networks/sanchonet/conway-genesis.json`
  - `components/common/src/main/resources/store/networks/sanchonet/config.json`
- Added Sanchonet to public-network checks in:
  - `BootstrapPhaseService`
  - `DRepDistService`
- Fixed committee reconstruction after `NO_CONFIDENCE` so the next `UPDATE_COMMITTEE` starts from an empty committee when appropriate.
- Fixed ratification boundary handling so the current epoch committee is used, but committee registrations/deregistrations are only considered if they existed before entering that epoch.


satran004 (approved):
LGTM. Thanks @Sotatek-HuyLe3a

PR #912, opened 2026-04-29
feat(core): add Cardano node health indicators

matiwinnetou:
## Summary

Two small additions to `components/core` that any yaci-store deployment benefits from — especially the use case of running yaci-store as a **standalone application** (the `applications/all` distribution, the docker images, the binary release) rather than embedded as a library in a larger service. Today yaci-store ships with **zero `HealthIndicator` implementations** anywhere in the repo; this PR adds the two most useful generic ones.

### 1. `NodeHealthIndicator` — liveness for the n2n / n2c connection

Drops into Spring Boot Actuator's `/actuator/health` alongside the built-in `db`, `diskSpace`, `ping` entries. Reports whether yaci-store is currently able to talk to a Cardano node and receive blocks — independent of how far behind the tip it is.

```jsonc
"nodeHealth": {
  "status": "UP",
  "details": {
    "connectionAlive": true,
    "receivingBlocks": true,
    "error": false,
    "syncStatus": "Connected"
  }
}
```

Status mapping:

| Condition | Status | `syncStatus` detail |
|---|---|---|
| Connected and receiving blocks | `UP` | `Connected` |
| Connected but no blocks flowing | `OUT_OF_SERVICE` | `Not receiving blocks` |
| Sync scheduled to stop | `OUT_OF_SERVICE` | `Scheduled to stop` |
| Connection lost / sync error | `DOWN` | `Connection lost or sync error` |
| Block fetcher not initialised yet | `UNKNOWN` | `Block fetcher not initialized` |

### 2. `NodeSyncIndicator` — readiness for chain-tip catchup

The other half of the picture. Reports `UP` only when the indexer is within 10 blocks of the chain tip (per the existing `SyncStatusService` definition); pods that are still catching up to the tip report `OUT_OF_SERVICE` so Kubernetes readiness probes / load balancers don't route traffic to them prematurely. Operators get `syncPercentage` and `timeSinceLastBlockMs` in the details for dashboards.

```jsonc
"nodeSync": {
  "status": "OUT_OF_SERVICE",
  "details": {
    "connectionAlive": true,
    "receivingBlocks": true,
    "error": false,
    "timeSinceLastBlockMs": 350,
    "syncPercentage": "42.50%",
    "syncStatus": "Syncing"
  }
}
```

## Why this belongs in core

- **Both indicators are pure passthroughs of `core.service.HealthService` / `core.service.SyncStatusService`** — they don't read any extension-specific state. Putting them in core makes them available to every yaci-store starter (utxo, blocks, transaction, ledger-state, governance, etc.) instead of any individual store/extension.
- **Read-only mode is handled correctly**: both indicators are gated with `@ConditionalOnBean(HealthService.class)`, so when `store.read-only-mode=true` is set (and `HealthService`, which is `@ReadOnly(false)`, is not created) the indicators silently drop out instead of failing on a missing dependency.

## Compatibility

- `components/common` already exports `spring-boot-starter-actuator` via `api`, so core inherits `HealthIndicator` on its compile classpath transitively. No `build.gradle` change needed.
- No public-API change. New classes only; no existing class touched.

## Test plan

- [x] `NodeHealthIndicatorTest` — 6 cases covering all status branches (UP, DOWN on connection lost, DOWN on error, OUT_OF_SERVICE for not-receiving-blocks, OUT_OF_SERVICE for scheduled-to-stop, UNKNOWN when block fetcher uninitialised)
- [x] `NodeSyncIndicatorTest` — 6 cases covering UP when fully synced, OUT_OF_SERVICE while syncing, OUT_OF_SERVICE for scheduled-to-stop, DOWN on connection lost, OUT_OF_SERVICE for not-receiving-blocks, UNKNOWN when block fetcher uninitialised
- [x] `./gradlew :components:core:test` green


components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/health/NodeSyncIndicator.java
```diff
@@ -0,0 +1,102 @@
...
+ * </ul>
+ *
+ * <p>Conditional on {@link HealthService}: in read-only mode
+ * ({@code store.read-only-mode=true}) neither {@code HealthService} nor
+ * {@code SyncStatusService} is created (both are {@code @ReadOnly(false)}),
+ * so this indicator is silently skipped instead of failing on a missing
+ * dependency.
+ */
+@Component("nodeSync")
+@ConditionalOnBean(HealthService.class)
```
ducpm2303, line 44: Should we also add the SyncStatusService ?

matiwinnetou: SyncStatusService is already available in the core: com.bloxbean.cardano.yaci.store.core.service.SyncStatusService

it was ported before I believe from admin module.

satran004:
@Sotatek-HuyLe3 Can you please check this PR before we merge to main?

Sotatek-HuyLe3a (approved):

Sotatek-HuyLe3a:
LGTM!

PR #925, opened 2026-05-08
[2-0-x] feat(docker): support multi-instance deployment via parameterized compose config

Sotatek-HuyLe3a:


satran004 (approved):
Thanks @Sotatek-HuyLe3a . I think we need the setup instruction in docs also.

PR #935, opened 2026-05-12
 Add withdrawable rewards API from indexed data (no N2C required)

Sotatek-HuyLe3a:
## Summary

  Adds DB-backed withdrawable reward support for stake accounts, including:

  - `GET /accounts/{address}/rewards/withdrawable`
  - fallback reward data for `/accounts/{stakeAddress}` when N2C local state access is unavailable
  - reward calculation based on spendable rewards after the latest withdrawal

  ## Motivation

  Non-N2C deployments cannot query stake account rewards through local state, so `/accounts/{stakeAddress}` may return missing or zero reward data even when AdaPot reward calculation data is available in the store.

  This change exposes withdrawable rewards directly from indexed AdaPot data and allows the account API to use that data once reward calculation has completed. 
  
  
##  Changes

  New endpoint:
  GET /api/v1/accounts/{address}/rewards/withdrawable

  Returns the withdrawable reward amount, current epoch, and max current slot for a stake address. The calculation sums rewards from reward, reward_rest, and instant_reward tables where spendable_epoch <= current_epoch, excluding
  rewards at or before the latest withdrawal slot.

##  Architecture:

  - Introduced StakeAccountRewardProvider interface in the account module —I that allows different modules to supply reward info for a stake address.
  - AdaPotStakeAccountRewardProvider implements this interface using the AdaPot reward storage, guarded by a check that reward calculation has completed at least once.
  - AccountService.getAccountInfo() now tries the local node state query first (N2C), then falls back to registered StakeAccountRewardProvider implementations. Previously it threw an exception when no local node was
  configured.
  - Added findWithdrawableRewardByAddress() to RewardStorageReader with a JOOQ implementation that correctly handles the withdrawal-as-reset-boundary semantics.


aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/api/account/service/AccountService.java
```diff
@@ -27,10 +30,18 @@
...
     private final LocalClientProviderManager localClientProviderManager;
+    private final List<StakeAccountRewardProvider> stakeAccountRewardProviders;
 
     public AccountService(@Nullable LocalClientProviderManager localClientProviderManager, EraStorage eraStorage) {
+        this(localClientProviderManager, eraStorage, List.of());
+    }
+
+    @Autowired
+    public AccountService(@Nullable LocalClientProviderManager localClientProviderManager, EraStorage eraStorage,
+                          List<StakeAccountRewardProvider> stakeAccountRewardProviders) {
```
satran004: Why list for StakeAccountRewardprovider here?

aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/api/account/service/AccountService.java
```diff
@@ -40,8 +51,22 @@ public AccountService(@Nullable LocalClientProviderManager localClientProviderMa
...
      */
     public Optional<StakeAccountRewardInfo> getAccountInfo(String stakeAddress) {
-        if (localClientProviderManager == null)
-            throw new IllegalStateException("LocalClientProvider is not initialized. Please check n2c configuration.");
+        var localAccountInfo = getAccountInfoFromLocalState(stakeAddress);
+        if (localAccountInfo.isPresent()) {
+            return localAccountInfo;
+        }
+
+        return stakeAccountRewardProviders.stream()
```
satran004: I think we should simplify this as there will be only one stakeAccountRewardProvider.

satran004 (approved):

PR #872, opened 2026-03-30
feat: implement Blockfrost Metadata API

ducpm2303:
# **Module 11: blockfrost-metadata (3 endpoints) — IMPLEMENTED**

---

## Implementation Notes

```

BFMetadataController
    @RequestMapping("${blockfrost.apiPrefix}/metadata/txs")
    @ConditionalOnExpression("${store.extensions.blockfrost.metadata.enabled:false}")
    ↓
BFMetadataService
    ├─ MetadataService (stores-api/metadata-api)         
    │       └─ TxMetadataStorageReader (stores/metadata)  
    │               └─ JOOQ / JPA → transaction_metadata table
    │
    └─ BFMetadataStorageReader (blockfrost-local)        
            └─ BFMetadataStorageReaderImpl (JOOQ)
    ↓
BFMetadataMapper (MapStruct abstract class)
    ├─ toJsonDto(TxMetadataLabel)
    │      
    └─ toCborDto(TxMetadataLabel)
    ↓
DTOs (Blockfrost format, @JsonNaming SnakeCaseStrategy)
    ├─ BFMetadataLabelDto  
    ├─ BFMetadataJsonDto   
    └─ BFMetadataCborDto   
```

## Status Checklist (per endpoint)

| Endpoint                              | Difficulty | Data    | Service Available                                    | Status |
| ------------------------------------- | ---------- | --------| ---------------------------------------------------- | ------ |
| GET /metadata/txs/labels              | MEDIUM     | PARTIAL | BFMetadataService.getLabels()                        | [x]    |
| GET /metadata/txs/labels/{label}      | EASY       | YES     | BFMetadataService.getMetadataByLabel()               | [x]    |
| GET /metadata/txs/labels/{label}/cbor | MEDIUM     | PARTIAL | BFMetadataService.getMetadataCborByLabel()           | [x]    |

---

## Open Issues / Known Limitations

* **`cip10` always `null`** — `BFMetadataLabelDto.cip10` is hardcoded to `null` in `BFMetadataStorageReaderImpl`. CIP-10 defines a community registry mapping numeric labels to human-readable names, but no integration with that registry is implemented.

* **Numeric label sorting workaround** — The `/labels` endpoint sorts by casting `label` to `BigDecimal` (`LABEL.cast(BigDecimal.class)`) rather than a native `BIGINT` cast. 

[...]

PR #865, opened 2026-03-25
feat: add bf governance apis

ducpm2303:
## Module: Blockfrost Governance APIs (17 endpoints)

Closes #864

---

## Summary

This PR includes fixes and improvements for the Blockfrost governance APIs under `/governance`, based on latest preprod runs:

- DRep amount fix with N2C connection
- Delegators query improvement with window function
- Proposal list optimization
- Protocol parameters returned as strings (matching Blockfrost)
- Proposal detail now returns anchor_url and anchor_hash

## Review guide

| Area | Files | What to review |
|---|---|---|
| Wiring | `BFGovernanceConfiguration`, `BFGovernanceProperties` | N2C configuration and feature flags |
| API surface | `BFGovernanceController`, `BFProposalsController`, DTOs | endpoint contracts |
| Mapping | `BFDRepMapper`, `BFProposalMapper` | Blockfrost field mapping |
| Query logic | `BFGovernanceStorageReaderImpl` | DRep/delegator/proposal queries, window functions |
| Reader models | `storage/impl/model/*` | query shapes for DRep, proposal, votes |

### Data flow

```mermaid
flowchart LR
    A[BFGovernanceController] --> B[BFGovernanceService]
    B --> C[BFGovernanceStorageReaderImpl]
    B --> D[BFDRepMapper]
    B --> E[BFProposalMapper]

    C --> F[(governance)]
    C --> G[(governance_aggr)]
    C --> H[(utxo)]
```

---

## Endpoints

| # | Endpoint | Status | Notes |
|---|---|---|---|
| 1 | `GET /governance/dreps` | ⚠️ Mismatch | ordering + special DReps not in list |
| 2 | `GET /governance/dreps/{drep_id}` | ✅ Match | amount now works via N2C |
| 3 | `GET /governance/dreps/{drep_id}/delegators` | ✅ Match | Added indexes: delegation_vote(address,slot), delegation_vote(drep_hash), address_utxo(owner_stake_addr) |
| 4 | `GET /governance/dreps/{drep_id}/updates` | ✅ Match | |
| 5 | `GET /governance/dreps/{drep_id}/votes` | ✅ Match | |
| 6 | `GET /governance/dreps/{drep_id}/metadata` | ⚠️ Mismatch | off-chain fields not indexed |
| 7 | `GET /governance/proposals` | ✅ Match | uses BFProposalListItemDto (summary only) |
[...]

PR #852, opened 2026-03-16
feat: bf scripts apis

ducpm2303:
## Module: Blockfrost Scripts APIs (7 endpoints)

Closes #851

---

## Summary

This PR implements the Blockfrost scripts module end-to-end:

- controller, service, storage reader, DTOs, mapper, and starter wiring
- all 7 scripts endpoints under `/scripts`
- cross-store JOOQ queries over `script`, `transaction`, and `datum` stores
- follow-up fixes from live comparison against preprod Blockfrost

### Visual status

```mermaid
pie showData
    title Live validation result (3 datasets × 53 checks)
    "Matched" : 153
    "Accepted sync-gap limitations" : 6
```

---

## Review guide

| Area | Files | What to review |
|---|---|---|
| Wiring | `BFScriptsConfiguration`, `BFAutoConfiguration`, `BFProperties`, starter/build files | module registration and feature-flag wiring |
| API surface | `BFScriptsController`, `BFScriptsService`, DTOs | endpoint contract and response shape |
| Mapping | `BFScriptsMapper`, `BFScriptsMapperDecorator` | Blockfrost field normalization and script type mapping |
| Query logic | `BFScriptsStorageReaderImpl` | script list ordering, type/CBOR/JSON selection, redeemer fee computation, datum lookup |
| Reader models | `storage/impl/model/*` | internal query result shapes |

### Data flow

```mermaid
flowchart LR
    A[BFScriptsController] --> B[BFScriptsService]
    B --> C[BFScriptsStorageReaderImpl]
    B --> D[BFScriptsMapper]

    C --> E[(script)]
    C --> F[(transaction)]
    C --> G[(datum)]
    C --> H[(epoch_param)]
```

---

## Endpoints

| # | Endpoint | Status | Notes |
|---|---|---|---|
| 1 | `GET /scripts` | ⚠️ Partial | ordering correct; partial overlap due to local sync gap (see accepted limitation) |
| 2 | `GET /scripts/{script_hash}` | ✅ Match | `script_hash`, `type`, `serialised_size` for all 4 script types |
| 3 | `GET /scripts/{script_hash}/json` | ✅ Match | native script JSON match; Plutus returns `null` |
| 4 | `GET /scripts/{script_hash}/cbor` | ✅ Match | Plutus CBOR hex match; native returns `null` |
| 5 | `GET /scripts/{script_hash}/redeemers` | ✅ Match | all fields including `fee`; validated across 54-row, 64-row, 57-row scripts |
[...]

PR #847, opened 2026-03-12
Blockfrost pools apis

ducpm2303:
## Module 7: Blockfrost Pool APIs (12 endpoints)

Closes #846

---

## Summary

This PR focuses on current Blockfrost parity mismatches for the pool module under `/pools`, based on latest preprod reruns.

- documents remaining mismatches only
- separates yaci-store limitations from upstream/preprod behavior gaps
- provides dataset-backed evidence and artifact links for each mismatch class

---

## Review guide

| Area | Files | What to review |
|---|---|---|
| Wiring | `BFPoolsConfiguration`, `BFAutoConfiguration`, `BFProperties`, starter/build files | module registration and feature flag wiring |
| API surface | `BFPoolsController`, `BFPoolsService`, DTOs | endpoint contract and response shape |
| Mapping | `BFPoolsMapper` | Blockfrost field mapping/normalization |
| Query logic | `BFPoolsStorageReaderImpl` | lifecycle-state selection, history/delegator/reward joins, relay extraction |
| Reader models | `storage/impl/model/*` | internal query shapes for pool, history, and stake info |

### Data flow

```mermaid
flowchart LR
    A[BFPoolsController] --> B[BFPoolsService]
    B --> C[BFPoolsStorageReaderImpl]
    B --> D[BFPoolsMapper]

    C --> E[(staking)]
    C --> F[(blocks)]
    C --> G[(adapot)]
    C --> H[(governance)]
```

---

## Endpoints

| # | Endpoint | Status | Notes |
|---|---|---|---|
| 1 | `GET /pools` | ✅ Match | |
| 2 | `GET /pools/extended` | ⚠️ Mismatch | stake semantics + off-chain metadata fields |
| 3 | `GET /pools/retired` | ⚠️ Mismatch | same-epoch tie ordering drift |
| 4 | `GET /pools/retiring` | ✅ Match | |
| 5 | `GET /pools/{pool_id}` | ⚠️ Mismatch | `calidus_key` + live/active size semantics |
| 6 | `GET /pools/{pool_id}/history` | ⚠️ Mismatch | tiny floating serialization tail on one dataset |
| 7 | `GET /pools/{pool_id}/metadata` | ⚠️ Mismatch | off-chain metadata unsupported |
| 8 | `GET /pools/{pool_id}/relays` | ⚠️ Mismatch | upstream IPv6 decode issue on affected pools |
| 9 | `GET /pools/{pool_id}/delegators` | ✅ Match | |
[...]

PR #958, opened 2026-05-21
[Do Not Merge] docs(blockfrost-adr): add per-module gap tracking docs for all 11 BF modules

ducpm2303:
Adds extensions/blockfrost/adr/{module}_gaps.md for each of the 11 Blockfrost compatibility modules. Each file captures:
- Endpoint-level match/mismatch status
- Open gaps categorised as accepted limitations or tracked issues
- Required index SQL (CONCURRENTLY IF NOT EXISTS)
- Configuration prerequisites
- Release readiness notes

Modules covered: epoch (#780), address (#784), asset (#780), block (#811), transaction (#818), account (#836), pools (#847), scripts (#852), governance (#865), network (#866), metadata (#872).

### DIFF
--- a/extensions/blockfrost/adr/asset_gaps.md
+++ b/extensions/blockfrost/adr/asset_gaps.md
@@ -6,7 +6,7 @@
 
 | Endpoint | Match | Notes |
 |----------|-------|-------|
-| `GET /assets` | ⚠️ | Pagination drift from phantom assets; slow without summary table |
+| `GET /assets` | ⚠️ | Slow without summary table |
 | `GET /assets/{asset}` | ⚠️ | `metadata` and `onchain_metadata` fields null |
 | `GET /assets/{asset}/history` | ✅ | |
 | `GET /assets/{asset}/txs` | ✅ | |
@@ -20,9 +20,6 @@
 
 - **Metadata fields null:** `metadata` and `onchain_metadata` are always `null` — no token
   registry or CIP-68 metadata ingestion is implemented. Requires a separate integration component.
-
-- **Phantom asset pagination drift:** 2 phantom assets in preprod data cause a 1-row shift
-  on paginated results. Root cause is a preprod data inconsistency, not a query bug.
 
 - **`asset_name` as hex:** Returned as hex substring of `unit` (chars 57+) to match Blockfrost
   convention. Intentional.
@@ -71,3 +68,5 @@ CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_transaction_tx_hash_tx_index
 
 No functional blockers. Performance of `GET /assets` is the main open item.
 Metadata fields gap is accepted (requires separate integration work).
+Phantom asset pagination drift (previously noted for 2 preprod assets) was resolved
+as part of the invalid-tx fix.
--- a/extensions/blockfrost/adr/asset_gaps.md
+++ b/extensions/blockfrost/adr/asset_gaps.md
@@ -1,0 +1,73 @@
+# Asset Module — Gaps & Open Items
+
+**PR:** #780 | **Status:** ✅ Merged | **Endpoints:** 7 / 7
+
+## Endpoint Status
+
+| Endpoint | Match | Notes |
+|----------|-------|-------|
+| `GET /assets` | ⚠️ | Pagination drift from phantom assets; slow without summary table |
+| `GET /assets/{asset}` | ⚠️ | `metadata` and `onchain_metadata` fields null |
+| `GET /assets/{asset}/history` | ✅ | |
+| `GET /assets/{asset}/txs` | ✅ | |
+| `GET /assets/{asset}/transactions` | ✅ | |
+| `GET /assets/{asset}/addresses` | ✅ | |
+| `GET /assets/policy/{policy_id}` | ✅ | |
+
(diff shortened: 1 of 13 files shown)


extensions/blockfrost/adr/asset_gaps.md
```diff
@@ -0,0 +1,73 @@
...
+| `GET /assets/policy/{policy_id}` | ✅ | |
+
+## Open Gaps
+
+### Accepted Limitations
+
+- **Metadata fields null:** `metadata` and `onchain_metadata` are always `null` — no token
+  registry or CIP-68 metadata ingestion is implemented. Requires a separate integration component.
+
+- **Phantom asset pagination drift:** 2 phantom assets in preprod data cause a 1-row shift
```
Sotatek-HuyLe3a: it seems to be fixed when we fixed invalid tx issue.

extensions/blockfrost/adr/epoch_gaps.md
```diff
@@ -0,0 +1,53 @@
...
+| `GET /epochs/{number}/stakes` | ⚠️ | `active_stake` is `null` when `store.adapot.enabled=false` |
+| `GET /epochs/{number}/stakes/{pool_id}` | ⚠️ | Same adapot dependency |
+| `GET /epochs/{number}/blocks` | ⚠️ | Slow without composite index (see below) |
+| `GET /epochs/{number}/blocks/{pool_id}` | ⚠️ | Same performance issue |
+
+## Open Gaps
+
+### Accepted Limitations
+
+- **`active_stake` null:** When `store.adapot.enabled=false`, stake fields return `null`.
```
Sotatek-HuyLe3a: I think we should not consider this as a gap, It feels more like a note when it depends on the adopt job.
There are many things that depend on the adopt job.

extensions/blockfrost/adr/metadata_gaps.md
```diff
@@ -0,0 +1,60 @@
...
+
+- **Raw JSON passthrough:** Uses `@JsonRawValue` for metadata content. Malformed stored metadata
+  produces structurally invalid JSON responses. Acceptable as the source data is already invalid.
+
+- **Numeric label sort workaround:** Labels cast to `BigDecimal` for sort ordering instead of
+  native `BIGINT`. Functionally equivalent but less efficient. Low priority.
+
+### Tracked Issues
+
+- **Duplicate rows on reconnect:** `GET /metadata/txs/labels` can return duplicate label rows
```
Sotatek-HuyLe3a: I think this is a very rare case, hard to reproduce? Do you think so, @ducpm2303 ?

extensions/blockfrost/adr/network_gaps.md
```diff
@@ -0,0 +1,57 @@
+# Network Module — Gaps & Open Items
+
+**PR:** #866 | **Status:** 🔄 Open PR | **Endpoints:** 4 / 4 (76 / 82 field matches)
+
+## Endpoint Status
+
+| Endpoint | Match | Notes |
+|----------|-------|-------|
+| `GET /` (root) | ⚠️ | `url`/`version` reflect local service; root hardcodes mainnet URL |
```
Sotatek-HuyLe3a, line 9: Yaci-store can return it using its own URL, right?

extensions/blockfrost/adr/network_gaps.md
```diff
@@ -0,0 +1,57 @@
...
+| `GET /genesis` | ✅ | Exact match on all 10 fields; sourced from bundled config |
+| `GET /network` | ⚠️ | `supply.locked` = 0 (stub); `stake.live` = 0 (structural gap) |
+| `GET /network/eras` | ⚠️ | Conway era end is `null` until next hardfork fires |
```
Sotatek-HuyLe3a, line 12: Can you clarify this gap a bit more?

extensions/blockfrost/adr/network_gaps.md
ducpm2303: added the note, will fix in in network PR

extensions/blockfrost/adr/network_gaps.md
ducpm2303: I will test this one again, seems like a probem when harkfork occurs

extensions/blockfrost/adr/asset_gaps.md
ducpm2303: yes, updated

extensions/blockfrost/adr/epoch_gaps.md
ducpm2303: updated

extensions/blockfrost/adr/metadata_gaps.md
ducpm2303: yeah, i can not reproduce this one again, but this issue still there, so in the future we can take a look 
https://github.com/bloxbean/yaci-store/issues/889

satran004: I think this is due to the UUID as the primary key, but the rollback should have removed the previous record first. Let's keep the issue open for now. Instead of a UUID, we can consider a deterministic unique key using the tx hash, metadata label, and unique number.

PR #960, opened 2026-05-23
 chore: Set JVM DNS cache TTL defaults for Yaci 0.4.4

satran004:
## Summary
  
  Pair the upstream Yaci 0.4.4 stale-DNS-cache fix ([bloxbean/yaci#163](https://github.com/bloxbean/yaci/pull/163)) with the JVM-level TTL flags it requires. Without these flags the JVM keeps returning a stale cached IP and the upstream re-resolve-on-retry logic never sees a fresh address — the fix appears to ship but does nothing in practice.

  ## Changes

  - **`bin/start.sh`, `bin/start.bat`** — prepend `-Dsun.net.inetaddr.ttl=60 -Dsun.net.inetaddr.negative.ttl=10` to `JAVA_OPTS`. Any user-exported `JAVA_OPTS` is preserved (appended after our defaults so the user wins on conflicting `-D` keys).
  - **`docker/config/env`** — ship the same flags via `JAVA_TOOL_OPTIONS` as an always-on baseline. `JDK_JAVA_OPTIONS` stays reserved for operator-tunable heap/GC flags (the JVM combines both — they do not overwrite each other unless they target the same `-D` key).
  - **`docker/config/env` (polyglot example)** — migrate from the broken `JDK_JAVA_OPTIONS=${JDK_JAVA_OPTIONS} -Dloader.path=…` self-append pattern (Compose `env_file` does not interpolate,
  and duplicate keys overwrite) to `LOADER_PATH=…`, which Spring Boot's `PropertiesLauncher` reads natively as an env var.
  - **Docs** — update `docs/pages/plugins/plugin-getting-started.mdx` and `docs/pages/tutorials/tracking-address-utxos.mdx` to reference the new `LOADER_PATH=` form for the Docker setup.

### DIFF
--- a/bin/start.bat
+++ b/bin/start.bat
@@ -9,6 +9,9 @@ REM ------------------------------------------------------------
 REM Save the current working directory
 SET CURRENT_DIR=%CD%
 
+REM Lower JVM DNS cache TTLs to avoid stale DNS cache.
+SET "JAVA_OPTS=-Dsun.net.inetaddr.ttl=60 -Dsun.net.inetaddr.negative.ttl=10 %JAVA_OPTS%"
+
 REM Uncomment the following line to enable polyglot (Python, JS) plugins support.
 :: set "JAVA_OPTS=%JAVA_OPTS% -Dloader.path=plugins,plugins/lib,plugins/ext-jars"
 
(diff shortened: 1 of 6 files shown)


Sotatek-HuyLe3a (approved):

satran004 (commented):
Thanks. LGTM

PR #963, opened 2026-05-28
fix: align Blockfrost V3 cost model output

ducpm2303:
## Summary

This PR fixes Blockfrost epoch protocol parameter parity for Plutus V3 cost model output.

When comparing local Blockfrost-compatible epoch APIs against Blockfrost mainnet, the `/epochs/{number}/parameters` response showed mismatches in the Plutus V3 cost model values. The local response was returning V3 cost model entries in an order that did not match the ledger/Blockfrost order, causing value-to-operation mismatches in the serialized response.

## Problem

- Blockfrost expects Plutus V3 cost model values to follow ledger order.
- Local output derived V3 cost model operation names in a different order.
- This caused many Plutus V3 cost model fields to have incorrect values when compared with Blockfrost.
- Governance deposit fields also needed Blockfrost-compatible string serialization.

## Fix

- Updated Plutus V3 cost model operation ordering to preserve the ledger-defined order.
- Reused the existing V3 cost model source instead of adding a Blockfrost-specific duplicate model.
- Updated Blockfrost epoch protocol parameter mapping to:
  - preserve raw cost model data when available
  - serialize `gov_action_deposit` as a string
  - serialize `drep_deposit` as a string
## Validation
- [x] testing on mainnet
- [ ] testing on preprod
- [ ] testing on the preview

### DIFF
--- a/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/util/PlutusOps.java
+++ b/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/util/PlutusOps.java
@@ -691,10 +691,9 @@ public class PlutusOps {
             keys = map2.keySet();
             V2_OPS = keys.stream().sorted().toList();
 
-            Map<String, Long> map3 = objectMapper.readValue(PLUTUS_V3_COSTS, new TypeReference<SortedMap<String, Long>>() {
+            Map<String, Long> map3 = objectMapper.readValue(PLUTUS_V3_COSTS, new TypeReference<LinkedHashMap<String, Long>>() {
             });
-            keys = map3.keySet();
-            V3_OPS = keys.stream().sorted().toList();
+            V3_OPS = map3.keySet().stream().toList();
 
         } catch (Exception e) {
             throw new RuntimeException(e);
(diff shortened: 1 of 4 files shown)


stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/util/PlutusOps.java
```diff
@@ -691,10 +691,9 @@ private static void initOps() {
             keys = map2.keySet();
             V2_OPS = keys.stream().sorted().toList();
 
-            Map<String, Long> map3 = objectMapper.readValue(PLUTUS_V3_COSTS, new TypeReference<SortedMap<String, Long>>() {
+            Map<String, Long> map3 = objectMapper.readValue(PLUTUS_V3_COSTS, new TypeReference<LinkedHashMap<String, Long>>() {
```
satran004, line 694: @ducpm2303 Why not the same change for V1 and V2? Also, after changing this, we need to verify a script transaction using Cardano Client Lib or a JS sdk

satran004: Also, V3 now has additional builtins (~350). We need to update the V3 map once those new built-ins are available in BF.

Sotatek-HuyLe3a: Basically, the PR looks good to me, just have the same question for V1, V2.

ducpm2303:
https://github.com/bloxbean/yaci-store/pull/969 
this PR include finxes of this one. close for now

PR #966, opened 2026-06-01
fix(governance): align DRep active_until with ledger expiry replay

Sotatek-HuyLe3a:
## Summary

  Fix DRep `active_until` calculation in `governance-aggr` to match Cardano ledger behavior.

  This PR replaces the previous snapshot-style adjustment with a ledger-style replay of the raw DRep expiry state. The replay accounts for dormant epoch counter timing, proposal flushes, DRep registrations, DRep
  updates, and DRep votes.

  ## Root Cause

  Before this change, `yaci-store` calculated `active_until` from the current snapshot state rather than replaying the ledger state transition that maintains raw `DRepState.drepExpiry`.

  The original implementation:

  - Calculated an `expiry` value first.
  - Used `gov_epoch_activity.dormant_epoch_count` at the left-boundary epoch to adjust `active_until`.
  - Treated the current dormant period as a value that could be subtracted from `expiry`.
  - Used proposal activity data as a coarse snapshot signal instead of replaying proposal flush and DRep interaction events in ledger order.

  This diverged from `cardano-ledger` because ledger does not derive `active_until` from a final snapshot formula. Ledger maintains raw `drepExpiry` state over time:

  - At epoch boundary, ledger updates the dormant counter before processing in-epoch events.
  - A proposal transaction flushes pending dormant epochs into stored DRep expiry.
  - A DRep registration/update/vote recomputes stored expiry using the current dormant counter.
  - Proposal presence for dormant detection is based on valid proposal state, equivalent to `currentEpoch <= gasExpiresAfter`.

  The original calculation could therefore be off by one or more epochs.

  ## Changes

  - Add ledger-style `active_until` replay in `DRepExpiryUtil`.
  - Replay DRep registration, DRep update, DRep vote, and proposal submission events in epoch/slot/tx/event order.
  - Increment dormant counter at epoch boundary before processing events in that epoch.
  - Use proposal status epochs with `ACTIVE` or valid `RATIFIED` status as non-dormant epochs.
  - Guard proposal-status epochs by proposal lifetime:
    `status_epoch <= proposal_epoch + gov_action_lifetime`.
  - Stop using `gov_epoch_activity.dormant_epoch_count` to calculate `active_until`.
  - Add service-level tests for valid and expired `RATIFIED` proposal-status handling.
  - Add regression tests for dormant counter replay, proposal flushes, and DRep interaction behavior.
  - Table `gov_epoch_activity` and `expiry` field in drep_dist are deprecated
  ## Test Plan

  ### Unit / Service Tests

  ```bash
  git diff --check main..HEAD

  ./gradlew :aggregates:governance-aggr:test \
    --tests com.bloxbean.cardano.yaci.store.governanceaggr.service.DRepExpiryServiceTest \
    --tests com.bloxbean.cardano.yaci.store.governanceaggr.processor.DRepExpiryUtilTest
```
  ### Network Verification

  After deploying/reindexing with this branch, run DRep active_until comparison against cardano-db-sync for:

  - [x] Sanchonet
  - [x] Preprod
  - [x] Preview 
  - [x] Mainnet 

  For each network, compare yaci-store.drep_dist.active_until against cardano-db-sync DRep distribution / expiry baseline across the available Conway-era epoch range.


Sotatek-HuyLe3a:
AI Suggestion: 
1. Two divergent definitions of "dormant epoch" now live in the same method. active_until uses the new proposal-status derivation (ACTIVE or RATIFIED, lifetime-guarded); expiry still uses
  gov_epoch_activity, whose producer GovEpochActivityProcessor:61 counts only ACTIVE and applies no lifetime guard. DRepExpiryService.calculateAndUpdateExpiryForEpoch now writes two columns
  from mutually inconsistent dormancy models in one statement. If the new definition is right, GovEpochActivityProcessor should be fixed too; if the legacy path is truly dead, drop it.

  2. The deprecation in the PR description isn't in the code. Nothing is annotated @Deprecated, there's no changelog/doc change, and expiry is still computed and persisted from the
  now-known-divergent path. Pick one: mark and document, or remove.

  3. DRepExpiryUtil.isEpochRangeDormant is now dead — zero callers after the import removal (including tests).

  4. findLastInteractions is fully redundant with the new findInteractionInfos. The last interaction is just max(epoch) over the list you already fetch, with the same drepActivity. That's ~55
  lines of duplicated jOOQ plus a second DB round-trip per 200-DRep batch that can be deleted.

  5. Within-transaction event ordering — please double-check priority(). You order PROPOSAL(0) → REGISTRATION(1) → INTERACTION(2). Ledger order inside a tx is flush → vote-expiry → CERTS, in
  both placements (Certs.hs applies both at the CERTS base case, i.e. before any cert; Ledger.hs applies both before the CERTS sub-rule behind
  hardforkConwayMoveWithdrawalsAndDRepChecksToLedgerRule). So votes should sort before registration/update certs, not after. Observable only when a DRep registers (or re-registers) and votes
  in the same tx under PV9 with a non-zero pending counter — ledger's Map.adjust is a no-op on a not-yet-registered credential, so the registration value wins; your replay lets the vote
  overwrite it with epoch + activity - counter. Narrow, but this PR is specifically about getting ordering right.

  6. event_index merges two different numbering spaces — drep_registration.cert_index and voting_procedure.idx — as a single tiebreaker. Harmless today (both INTERACTION branches compute the
  same value), but misleading as a sort key.

  7. Test gaps.
  - The Postgres/MySQL branch of findNonDormantProposalStatusEpochs — the production path — has no coverage. The H2 and SQL branches compute the lifetime guard differently (Java vs a
  JSON-extract cast) with different NULL behaviour; only H2 is exercised.
  - activeUntilTreatsExpiredProposalStatusEpochAsDormant doesn't test what its name says. The epoch-495 row is excluded by the status IN (ACTIVE, RATIFIED) filter, not by the lifetime guard —
  I traced it, the guard never fires in any of the three tests. Add a case with an ACTIVE/RATIFIED row beyond proposal_epoch + gov_action_lifetime.
  - DRepExpiryServiceTest hand-rolls DDL instead of running the module's H2 migration → silent schema drift. And DB_CLOSE_DELAY=-1 with a fresh UUID database per test leaks one in-memory DB
  per test for the JVM lifetime; use DB_CLOSE_DELAY=0 or close in @AfterEach.

satran004 (approved):
Thanks @huylevt

Issue #781, opened 2026-01-29
[BUG] Epoch totalFees mismatch


ducpm2303:
Fees mismatch for epoch,  

The `total_fees` of `epoch` table calulated in `epoch_aggr` and mismatch with blockforst api.
The correct fees can calculate by the first query bebow

```
select sum(b.total_fees) from block b where b.epoch = ?  -- correct
select e.total_fees from epoch e where e."number" = ? -- mismatch
```

`adapot` table has fee column which is getting calculated during epoch transition. So, we may be able to use that for all previous epochs. For current epoch, we may need to make the query in transaction table.

PR #969, opened 2026-06-04
Fix/bf epoch parity

ducpm2303:
# fix/bf-epoch-parity — Epoch API Parity with Blockfrost

## What

Fixes parity gaps in the Blockfrost-compatible `/epochs` API, verified against the real Blockfrost
API on **preprod** (epoch 291) and **preview** (epoch 1316).

---

###  `start_time` / `end_time` return block timestamps instead of slot boundaries

**`BFEpochService.java`**

Blockfrost returns theoretical epoch boundaries: `genesisTime + epochNumber × epochLength`.
Yaci was returning actual first/last block timestamps from the DB, which can differ by minutes.

Added `applyTheoreticalEpochTimes()` using `GenesisConfig.getStartTime()` and `getEpochLength()`.

```
Before: start_time = 1779767278  (first block timestamp)
After:  start_time = 1779753600  (genesisTime + 291 × 432000) ✅
```

---

### BUG-3 — `/epochs/{epoch}/previous` returns epochs in descending order

**`BFEpochStorageReaderImpl.java`**

The query used `ORDER BY DESC`, returning `[290, 289, … 191]`.
Blockfrost returns ascending `[191, 192, … 290]`.

Rewrote `findPreviousEpochs()` with a subquery: fetch the N epochs DESC (correct window),
then wrap with `ORDER BY ASC`.

```
Before: [290, 289, … 191]
After:  [191, 192, … 290] ✅
```

---

###  Closed #963  `cost_models` uses numeric index keys instead of named op keys

**`PlutusOps.java`** · **`BFProtocolParamMapperDecorator.java`**

`PlutusOps.java` only had 166/175/251 op names. With Conway-era params (332/332/350 ops),
unrecognised positions fell back to numeric keys (`"000": 100788`).

- Updated `PLUTUS_V1/V2/V3_COSTS` with the full 332/332/350 op names. Order matches DB storage
  order (same as Blockfrost `cost_models_raw` positional order).
- `BFProtocolParamMapperDecorator` was rebuilding `cost_models_raw` from an alphabetical `TreeMap`,
  producing wrong positional order. Fixed to pass `protocolParamsDto.getCostModelsRaw()` directly,
  which `DomainMapperDecorator` already builds correctly from the DB array.

```
Before: 166 keys, "000": 100788         → 2031 diffs
After:  332 keys, "addInteger-...": 100788, raw array matches BF → 0 diffs ✅
```

---

### — `/epochs/{epoch}/stakes` not sorted by address

**`BFEpochService.java`**

Blockfrost returns stakes sorted by `stake_address`. Yaci returned DB insertion order,
causing all items to mismatch when comparing page-by-page.

Added `.sorted(Comparator.comparing(EpochStake::getAddress))` in `getEpochStakes()` and
`getEpochStakesByPool()` after retrieval, keeping the shared repository method unchanged.

---

### — `min_utxo` incorrect on Babbage+ eras

**`BFProtocolParamMapperDecorator.java`**

`min_utxo` was only set from `coinsPerUtxoSize` when `minUtxo == null`. On Babbage+ the legacy
`minUtxo` DB field is populated, so the guard was never true — leaving `min_utxo` as the old
Shelley value.

Removed the null guard. `coinsPerUtxoSize` is now always authoritative for both
`coins_per_utxo_word` and `min_utxo` when present.

---

### TYPE — `gov_action_deposit` / `drep_deposit` serialized as numbers instead of strings

**`BFProtocolParamsDto.java`** · **`BFProtocolParamMapperDecorator.java`**

Blockfrost returns these as JSON strings. The DTO had them as `BigInteger`, causing Jackson to
serialize as plain numbers.

Changed field types to `String` in the DTO. Added `.toString()` conversion in the decorator.

```
Before: "gov_action_deposit": 100000000000
After:  "gov_action_deposit": "100000000000" ✅
```

---

## Out of Scope

| | Description |
|---|---|
| BUG-2 | `fees` near-zero — [Issue #781](https://github.com/bloxbean/yaci-store/issues/781) |
| BUG-4 | `nonce` always null in `/parameters` |
| DATA | Stake snapshot data diverges from Blockfrost |

Tested on:

- [ ] mainnet 
- [x] preprod
- [x] preview


extensions/blockfrost/epoch/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/epoch/mapper/BFProtocolParamMapperDecorator.java
```diff
@@ -21,35 +18,30 @@ public BFProtocolParamsDto toBFProtocolParamsDto(ProtocolParamsDto protocolParam
...
-            Map<String, List<Long>> costModelsRaw =  protocolParamsDto.getCostModels().entrySet().stream()
-                    .collect(Collectors.toMap(
-                            Map.Entry::getKey,
-                            entry -> entry.getValue().entrySet().stream()
-                                    .map(Map.Entry::getValue)
-                                    .collect(Collectors.toList())
-                    ));
-            bfProtocolParamsDto.setCostModelsRaw(costModelsRaw);
+        // cost_models: named keys (alphabetical TreeMap) from DomainMapperDecorator — pass through.
+        // cost_models_raw: original positional array already built from DB in DomainMapperDecorator — use directly.
```
Sotatek-HuyLe3a: these comment lines may be more detailed than needed for this mapper, since the logic here is just passing through values already prepared upstream. Could we
remove it or simplify it?

extensions/blockfrost/epoch/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/epoch/service/BFEpochService.java
```diff
@@ -96,6 +102,7 @@ public List<BFEpochStakePoolDto> getEpochStakesByPool(int number, String poolId,
         String poolHash = normalizePoolHash(poolId);
         List<EpochStake> stakes = stakeStorageReader.getAllActiveStakesByEpochAndPool(number, poolHash, page, count);
         return stakes.stream()
+                .sorted(Comparator.comparing(EpochStake::getAddress))
```
Sotatek-HuyLe3a: The above repository queries page without an ORDER BY, so sorting the returned page does not guarantee global `stake_address` order, right?

Sotatek-HuyLe3a:
@ducpm2303 ,2 comments , otherwise look good to me

extensions/blockfrost/epoch/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/epoch/mapper/BFProtocolParamMapperDecorator.java
Sotatek-HuyLe3a: Fixed

extensions/blockfrost/epoch/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/epoch/service/BFEpochService.java
Sotatek-HuyLe3a: Actually the data in BF  isn't sorted alphabetically. It seems to follow the ordering from db-sync, so we need rollback this change, and accept that the order problem is a gap

satran004 (approved):

PR #972, opened 2026-06-04
[backport] Enable both submit-api and ogimios service for yaci-store all profile, so that controllers are available in YS all native image

Sotatek-HuyLe3a:
check #965


satran004 (approved):
Thanks @Sotatek-HuyLe3a

PR #976, opened 2026-06-06
Fix/blockfrost address compat

ducpm2303:
# Blockfrost address module: compatibility fixes + whale-scale hardening

Brings `/api/v1/blockfrost/addresses/*` into field- and ordering-parity with official
Blockfrost, and bounds worst-case latency for very high-volume ("whale") addresses.
Verified against official Blockfrost on **preprod, preview, and mainnet**.

### 1. Include `tx_index` in address UTXO responses
- **What:** add `tx_index` to `BFAddressUtxoDTO` and populate it in the mapper for
  `/addresses/{address}/utxos` and `/utxos/{asset}`.
- **Why:** Blockfrost returns `tx_index` per UTXO; Yaci omitted it, so payloads diverged.

### 2. Derive `data_hash` from inline datum
- **What:** in `BFAddressUtxoMapperDecorator`, compute `data_hash = blake2b-256(inline_datum)`
  at read time when the stored `data_hash` is null.
- **Why:** the indexer only persists `data_hash` for hash-referenced datums; for inline-datum
  outputs the column is null, while Blockfrost reports the hash of the inline datum bytes.
  Computed at read time, so no reindex is required.

### 3. Deterministic UTXO ordering in the utxo store
- **What:** add a `(slot, tx_hash, output_index)` tiebreak (`utxoSort`) to the jOOQ
  asset/credential UTXO queries in `UtxoStorageReaderImpl`.
- **Why:** slot-only ordering left UTXOs in the same slot in arbitrary, unstable order across
  calls. 

### 4. Blockfrost-compatible `/utxos` ordering — `(slot, tx_index, output_index)`
- **What:** serve `/addresses/{address}/utxos` and `/utxos/{asset}` from a new
  `fetchAddressUtxos` query in `BFAddressStorageReaderImpl` that joins `transaction` to source
  `tx_index` and orders by `(slot, tx_index, output_index)`.
- **Why:** Blockfrost orders an address' UTXOs by on-chain position. `tx_hash` order
  ≠ `tx_index` order, so an address holding unspent UTXOs from multiple transactions in the
  same slot came back in the wrong order. `tx_index` is not stored on `address_utxo`, and the
  utxo-store module cannot depend on the transaction module, so the ordering is applied in the
  Blockfrost layer. The plan is fully index-driven (owner index → `tx_input` PK anti-join →
  `transaction` PK join), with no sequential scans, and the inner join drops no rows.

### 5. Two-phase `/transactions` + bounded whale queries
- **What:** restructure `findAddressTransactions` into two phases — (1) page the distinct tx
  hashes from the address' combined output/spend set **without** joining `transaction`, then
  (2) resolve `tx_index`/`block_time` and apply the exact `(block, tx_index)` ordering for only
  the paged rows. Add a query timeout to the unbounded full-history queries (`/transactions`,
  `/total`, `/utxos`, balance).
- **Why:** the previous single query forced sequential scans over `transaction` (~51M) and
  `tx_input` (~336M) and sorted the entire union before applying the page limit. The two-phase
  form keeps the spent side on the `tx_input` PK index (EXPLAIN cost 157M → 35M) and makes
  normal/medium addresses fast. For extreme whales the full-history aggregation
  (`ORDER BY max(block)`) is irreducible on the read path, so the timeout converts a
  pool-exhausting hang into a fast, bounded per-request failure.

### 6. Refactor / cleanup
- Add the query-timeout constant to `common.Constants.QUERY_TIMEOUT_SECONDS` so it is shared.
- Remove verbose comments.

## Performance summary

| Endpoint | Normal address | Whale address |
|---|---|---|
| `/transactions` | fast (~60–130 ms) | bounded 500 @ timeout |
| `/utxos`, `/utxos/{asset}` | fast (~40–65 ms) | bounded 500 @ timeout |

### Per-network results

Measured Yaci vs official Blockfrost (Yaci is consistently faster on normal addresses;
Blockfrost responses are network round-trips, shown for reference).

| Network | `/transactions` (normal) | `/utxos` ordering | Yaci latency | Blockfrost latency | Whale endpoints |
|---|---|---|---|---|---|
| **mainnet** | OK, 0 diffs | matches (3 same-slot trigger addresses) | sub-second | ~0.4–1.4 s | bounded 500 @ 15 s |
| **preprod** | OK, 0 diffs | matches (2 trigger addresses) | ~40–125 ms | ~0.37–1.4 s | n/a (no whale-scale addresses) |
| **preview** | OK, 0 diffs | matches | ~40–65 ms | ~0.37–0.78 s | n/a |

Only diffs remaining on any network are the accepted `/extended` CIP-68 fields (see below).
On mainnet, very high-volume addresses hit the bounded `/transactions`, `/utxos`, and balance
500; all other addresses respond normally.

## Known limitations / follow-ups
- Whale `/transactions`, `/utxos`, and the balance endpoints (`/addresses`, `/extended`) return
  a bounded 500. A complete fix requires a writer-maintained, per-`(address, tx)` index carrying
  `tx_index`, indexed on `(address, block, tx_index)`, so paging becomes O(page) instead of
  O(history). That is a writer-side change (e.g. the `address_tx_amount` aggregate, currently not
  populated, or a slim dedicated table) and is out of scope for this read-path PR.

### DIFF
--- a/extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/BFAddressStorageReader.java
+++ b/extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/BFAddressStorageReader.java
@@ -2,6 +2,7 @@ package com.bloxbean.cardano.yaci.store.blockfrost.address.storage;
 
 import com.bloxbean.cardano.yaci.store.blockfrost.address.dto.BFAddressTransactionDTO;
 import com.bloxbean.cardano.yaci.store.blockfrost.address.storage.impl.model.BFAddressTotal;
+import com.bloxbean.cardano.yaci.store.common.domain.AddressUtxo;
 import com.bloxbean.cardano.yaci.store.common.model.Order;
 
 import java.util.List;
@@ -13,6 +14,16 @@ public interface BFAddressStorageReader {
     List<String> findTxHashesByAddress(String address, int page, int count, Order order);
 
     List<BFAddressTransactionDTO> findAddressTransactions(String address, int page, int count, Order order, String from, String to);
+
+    /**
+     * Unspent UTXOs for an address in Blockfrost-compatible on-chain order (slot, tx_index, output_index).
+     */
+    List<AddressUtxo> findAddressUtxos(String address, int page, int count, Order order);
+
+    /**
+     * Unspent UTXOs for an address holding a given asset unit, in Blockfrost-compatible on-chain order.
+     */
+    List<AddressUtxo> findAddressUtxosForAsset(String address, String unit, int page, int count, Order order);
 
     Optional<BFAddressTotal> getAddressTotal(String address);
 
--- a/extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/impl/BFAddressStorageReaderImpl.java
+++ b/extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/impl/BFAddressStorageReaderImpl.java
@@ -14,6 +14,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.jooq.Condition;
 import org.jooq.DSLContext;
 import org.jooq.Field;
+import org.jooq.Record1;
 import org.jooq.Record2;
 import org.jooq.Record4;
 import org.jooq.Result;
@@ -98,9 +99,15 @@ public class BFAddressStorageReaderImpl implements BFAddressStorageReader {
         SortField<?> pageBlockOrder = order == Order.desc ? pageBlock.desc() : pageBlock.asc();
         SortField<?> pageHashOrder = order == Order.desc ? combinedTxHash.desc() : combinedTxHash.asc();
 
+        Condition pageCondition = buildBlockRangeCondition(combinedBlock, fromRef, toRef);
+        List<String> boundaryViolators = fetchBoundaryTxIndexViolators(combinedTx, combinedBlock, fromRef, toRef);
+        if (!boundaryViolators.isEmpty()) {
+            pageCondition = pageCondition.and(combinedTxHash.notIn(boundaryViolators));
+        }
+
         List<String> pagedTxHashes = dsl.select(combinedTxHash, pageBlock)
                 .from(combinedTx)
-                .where(buildBlockRangeCondition(combinedBlock, fromRef, toRef))
+                .where(pageCondition)
                 .groupBy(combinedTxHash)
                 .orderBy(pageBlockOrder, pageHashOrder)
                 .limit(count)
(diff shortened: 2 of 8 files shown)


extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/BFAddressStorageReader.java
```diff
@@ -14,6 +15,16 @@ public interface BFAddressStorageReader {
 
     List<BFAddressTransactionDTO> findAddressTransactions(String address, int page, int count, Order order, String from, String to);
 
+    /**
+     * Unspent UTXOs for an address in Blockfrost-compatible on-chain order (slot, tx_index, output_index).
+     */
+    List<AddressUtxo> findAddressUtxos(String address, int page, int count, Order order);
```
satran004, line 21: @ducpm2303 Can't we use both these new methods from `UtxoStorageReader` ? We can update order logic to use (slot, tx_index, output_index) instead of  (slot, tx_index, output_index) in core reader impl. 

@Sotatek-HuyLe3a Do you see any risk if we change the ordering logic in core reader?

huylevt: `tx_index` is not stored in `address_utxo`, so core UTXO would need to join the transaction table or depend on the transaction store. That changes the module boundary and affects all core UTXO callers, while this
  ordering requirement is Blockfrost-specific.

  I think I prefer to keep core ordering deterministic as `(slot, tx_hash, output_index)` and keep Blockfrost ordering `(slot, tx_index, output_index)` in the Blockfrost address reader

satran004: @huylevt Thanks. Then it makes sense to keep it separate.

What is the performance impact of this join with the `transaction` table? Does it add an extra join just for sorting? If we are already joining the `transaction` table to fetch other columns in the query, then I think it's fine. However, if the join is only needed for sorting, we may need to rethink the approach and just add a known limitation.

Later, we can probably plan to add `tx_index` to `address_utxo`.

satran004: Also, it seems `tx_index` is now deprecated in BF api response https://docs.blockfrost.io/#tag/cardano--addresses/GET/addresses/{address}/utxos/{asset} . So we can go with existing implementation in PR, but add a note to optimize it later by adding tx_index field in `address_utxo` if performance impact is significant for big accounts.

extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/impl/BFAddressStorageReaderImpl.java
```diff
@@ -83,34 +88,99 @@ public List<String> findTxHashesByAddress(String address, int page, int count, O
...
+        Field<Long> pageBlock = DSL.max(combinedBlock).as("blockHeight");
+        SortField<?> pageBlockOrder = order == Order.desc ? pageBlock.desc() : pageBlock.asc();
+        SortField<?> pageHashOrder = order == Order.desc ? combinedTxHash.desc() : combinedTxHash.asc();
+
+        List<String> pagedTxHashes = dsl.select(combinedTxHash, pageBlock)
+                .from(combinedTx)
+                .where(buildBlockRangeCondition(combinedBlock, fromRef, toRef))
+                .groupBy(combinedTxHash)
+                .orderBy(pageBlockOrder, pageHashOrder)
+                .limit(count)
```
huylevt, line 106: we are applying limit/offset here, and order by pageBlock, and tx_hash, then in the final query, we select in tx hash list, and order by tx_index. 
Will this affect the final result? And can we join/resolve `transaction.tx_index` before applying `limit/offset`?

extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/BFAddressStorageReader.java
huylevt: yes, I think the best / long term way is adding tx_index field in address_utxo.So  we don't need do join for each request

ducpm2303: @satran004  Thanks for catching this, You're right to question the join. so I measured it on a real mainnet whale (~1.7M UTXO rows) using a query plan (no execution, zero load on the DB).

it turns out the join was only there to sort by `tx_index`, and its cost is basically zero: `12,741,727` with the join vs `12,741,176 `without — a 0.004% difference.
Both versions spend all their time on the same owner_addr lookup, so the join was never the bottleneck.

But since Blockfrost dropped tx_index from the /utxos response anyway, sorting by it doesn't help the client. So I [removed](https://github.com/bloxbean/yaci-store/pull/976/changes/b040aba817d809e6b505bbc703717ea44995fdb4) the join and sort by (slot, tx_hash, output_index) instead .
Same results, stable pagination, and one less table dependency.

extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/impl/BFAddressStorageReaderImpl.java
ducpm2303: Checked this on live `mainnet`, no transaction is ever lost or duplicated. Paging is done by `(block, tx_hash),` and since every tx hash is unique, each page is a clean, separate slice; the second step just re-sorts that page by `tx_index`.

The only quirk: if one block has several of the address's txs and lands right on a page boundary, the split is by tx_hash instead of tx_index . So only the hand-off between those two pages may differ from exact on-chain order (each page is still correct on its own).

I kept it this way on purpose: sorting strictly by `tx_index` first would need a lookup for every tx of the address, which is much slower on big accounts. Documented as a known limitation; the long-term fix is storing tx_index on `address_utxo`.

extensions/blockfrost/address/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/address/storage/BFAddressStorageReader.java
satran004: Thanks @ducpm2303

Sotatek-HuyLe3a (approved):

PR #981, opened 2026-06-09
feat(blockfrost): default all sub-modules to enabled when parent gate is active

huylevt:
- Add top-level gate store.extensions.blockfrost.enabled on BFAutoConfiguration
  - Change all 9 sub-module defaults from false to true in BFProperties
  - Set matchIfMissing=true on all BF*Configuration classes
  - Update @ConditionalOnExpression defaults to :true in all controllers
  - Add blockfrost Spring profile (application-blockfrost.properties)
  - Enable epoch-aggr in ledger-state profile (required by BF epoch endpoints)
  - Add blockfrost profile option to docker/config/env
  
 BREAKING CHANGE: Users who previously set individual sub-module properties
  (e.g. store.extensions.blockfrost.metadata.enabled=true) must now also set
  store.extensions.blockfrost.enabled=true or activate the blockfrost profile.

### DIFF
--- a/config/application-blockfrost.properties
+++ b/config/application-blockfrost.properties
@@ -1,0 +1,21 @@
+#######################################################################################
+# Blockfrost Profile Configuration
+# Activate this profile to enable all Blockfrost-compatible API endpoints.
+# Usage: spring.profiles.active=ledger-state,blockfrost
+#######################################################################################
+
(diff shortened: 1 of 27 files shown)


satran004 (approved):
Thanks @huylevt . LGTM

PR #984, opened 2026-06-10
[Blockfrost] Fix/blockfrost asset ordering

ducpm2303:
# Fix Blockfrost asset endpoints: ordering and burn sign

This PR fixes four places where the Blockfrost **asset** endpoints returned the right *data* but in the
wrong *order* (plus one where a burn amount had the wrong sign). Three of the four share one root cause:
Blockfrost orders rows within a block by **`tx_index`**, and we were ordering by `tx_hash`. On normal assets
nothing looks wrong; on assets with lots of activity in the same block, almost every row ends up out of place.

Everything below was checked against the **live Blockfrost API** on preprod, preview and mainnet. After the
fixes, every network passes the comparison with only one known difference left (asset metadata, which we don't
ingest yet).

## Commits

| Commit | Endpoint | What it fixes |
|--------|----------|---------------|
| `0cc9d40b` | `/txs`, `/transactions` | Make the slow query fast (slot-window) without changing the result order |
| `cd0f007f` | `/history` | Burn amounts had the wrong sign |
| `9b658442` | `/history` | Events in the same block came back in the wrong order |
| `87ac58e6` | `/addresses` | Holders came back in the wrong order (and `desc` was wrong in a subtle way) |
| `3d7b9e46` | `/assets/policy` | Assets first minted in the same block came back in the wrong order |
| `a9debea4` | `/assets/policy` | Make huge policies fast (12–16 s with timeouts → 2.5 s cold / sub-second warm) |

---

## 1. `/history` — burn amount sign (`cd0f007f`)

A burn was returned as a **positive** number; the live API returns it **negative**. The database already stores
the burn negative — we were flipping it with `.abs()`.

```
GET /assets/{asset}/history   (a burn event)

            amount
Before      "500"      (we called .abs())
After       "-500"
Live BF     "-500"     now matches
```

The Blockfrost OpenAPI *example* actually shows a positive burn, so this is a deliberate choice to match the
**live API** rather than the doc example. Direction is also carried by the `action` field (`minted`/`burned`).

---

## 2. `/history` — order of events in the same block (`9b658442`)

When several mint/burn events land in the same block, Blockfrost orders them by `tx_index`. We ordered by
`tx_hash`, so the rows came out alphabetically by hash instead.

Real example — three events in one block (slot `41732728`) of a preview token:

```
tx_index   tx_hash
   0        d2787e7f…   (mint)
   1        c64baaa8…   (burn)
   3        049f45e7…   (mint)

Before (order by tx_hash):     049f45e7, c64baaa8, d2787e7f
After  (order by tx_index):    d2787e7f, c64baaa8, 049f45e7
Live BF:                       d2787e7f, c64baaa8, 049f45e7   
```

The `assets` table doesn't store `tx_index`, so the fix joins `transaction` to get it. To avoid joining and
sorting a whole 100k-event history just to return one page, the Postgres path narrows the work to the page's
slot range first (same trick the `/txs` query already uses).

---

## 3. `/addresses` — order of holders (`87ac58e6`)

We ordered holders by `(first_seen_slot, address)`. Two problems:

**(a)** Within the same slot, Blockfrost again uses `tx_index`, not the address string.

Real example — three holders that first appear in slot `42910881`:

```
tx_index   address
   2        addr…qzz8
   3        addr…qpe7
   5        addr…qrh5

Before (order by address):     qpe7, qrh5, qzz8
After  (order by tx_index):    qzz8, qpe7, qrh5
Live BF (asc):                 qzz8, qpe7, qrh5   
```

**(b)** `desc` is **not** just `asc` reversed. Blockfrost keys `asc` on each holder's **earliest** UTXO and
`desc` on each holder's **latest** UTXO. A holder who first received the token long ago but moved it recently
sits near the bottom in `asc` and near the top in `desc`. We confirmed this against the live API:

```
order=desc, first page

Before (reverse of asc):   wrong — 44 of 100 rows out of place
After  (key on latest utxo): matches live BF, every page
```

So the fix keys each holder on its earliest UTXO for `asc` and its latest UTXO for `desc`
(`slot, tx_index, output_index`), with the address as a final tie-break.

> Note: this endpoint is still **slow** on assets with many holders — it has to scan every holder UTXO. That
> is a separate performance task (a holder summary table); this PR only fixes the order.

---

## 4. `/assets/policy/{policy_id}` — order of assets (`3d7b9e46`)

Same story as #2, but for the first mint of each asset under a policy. We ranked the first mint by
`(slot, tx_hash)`; Blockfrost uses `tx_index`. This was invisible until we tested a large mainnet policy where
many assets were first minted in the same block.

Real example — first page of a mainnet policy (`…8484`), asset names shortened:

```
Before (Yaci):   757fae848430, 843933303235, 373930303739, 8468764d494e, 313438353635, 848432343234, …
After / Live BF: 757fae848430, 843933303235, 848432343234, 313438353635, 373930303739, 8468764d494e, …
```

Same assets, same quantities — only the order differed (98 of 100 rows on page 1). Fix: join `transaction`,
look only at mint rows, and order by `(slot, tx_index, unit)`.

---

## 5. `/txs` & `/transactions` — performance (`0cc9d40b`)

Not a correctness fix, but it's the query the `/history` fix is built on, so it's included here. The old query
joined `transaction` for *every* transaction of the asset (tens of thousands) just to sort and keep 100 rows.
The new query figures out the page's slot range first and only joins the transactions in that range. **The
returned order is unchanged.**

```
                preprod (deep unit)
Before          ~11.5 s
After           ~0.3 s
```

---

## 6. `/assets/policy` — performance (`a9debea4`)

The ordering fix (#4) made `/policy` correct but it was still slow on big policies: the query joined
`transaction` for *every* mint row of the policy and summed the quantity of *every* unit, just to return
100 rows. On a mainnet policy with 292k assets that was 12–16 s, and asc page 1 hit the 30 s client timeout.

Two changes, both in the Postgres path of `findAssetsByPolicy`:

1. **The first-mint pass now rides an existing index.** An asset's `unit` starts with its policy id, so
   `unit >= policy AND unit < policy || 'g'` turns the scan into an index-only scan on the partial mint
   index that's already there. No new index needed.

   ```
   first-mint pass on the 292k-asset policy:   5.5 s  →  0.6 s
   ```

2. **The slot-window trick** (same as `/history` and `/txs`) so the `transaction` join and the quantity
   sums only run for the ~100 units on the requested page, not the whole policy.

The returned order is unchanged — re-verified against live Blockfrost (asc/desc, pages 1–2, all three networks).

```
/assets/policy, measured after deploy        Yaci before   Yaci after        BF
mainnet (292k-asset policy), asc page 1      15.2 s        2.5 s (cold)      3.8 s
mainnet, warm pages                          12–16 s       0.8–1.3 s         0.4–4.1 s
preprod                                      0.4–3.1 s     53–85 ms          459–556 ms
preview                                      0.7–6.7 s     99–259 ms         351–616 ms
```

Yaci now beats Blockfrost on `/policy` everywhere, including the heaviest policy's first page.
Policies that aren't a standard 56-char id fall back to the previous query.

---

## Current performance 

Latency on the heaviest unit per network. "warm" is a repeated call; the **first call** of each endpoint pays a
cold aggregation cost, shown in parentheses. These are the worst-case units we could find, not typical ones.
**Yaci** = this store, **BF** = Blockfrost on the same unit.

| Endpoint | preview Yaci | preview BF | preprod Yaci | preprod BF | mainnet Yaci | mainnet BF |
|----------|-----------|-----------|-----------|-----------|-----------|-----------|
| `/assets/{asset}` | **282 ms** | 1.5 s | **147 ms** | 0.9 s | 3.7 s | 2.5 s |
| `/history` | 391–413 ms | 0.74–0.91 s | **169–182 ms** | 0.83–1.17 s | 3.9–4.1 s | 2.9–3.5 s |
| `/txs` | ~1.4 s | 0.29–0.71 s | **187–213 ms** | 0.41–0.72 s | 6.2–6.4 s | 0.36–0.51 s |
| `/transactions` | ~1.45 s | 0.56–0.81 s | ~200 ms | 0.51–0.92 s | 6.2–6.4 s | 0.55–1.6 s |
| `/addresses` | 2.1–2.5 s | 0.57–2.34 s | 0.47–0.59 s | 0.42–0.85 s | 8.8–10.8 s | 0.41–2.5 s |
| `/assets/policy` | **61–99 ms** | 0.44–0.53 s | 55–112 ms | 0.46–0.98 s | **0.77–1.05 s** | 0.84–3.7 s |

### A normal asset (for contrast)

The numbers above are the worst units we could find. A typical mainnet asset (a 2-event token, a few holders)
is fast on every endpoint and beats Blockfrost on most:

| Endpoint | Yaci | Blockfrost |
|----------|------|------------|
| `/assets/{asset}` | 221 ms | 640 ms |
| `/history` | 206–280 ms | 448–1032 ms |
| `/txs` | 265–471 ms | 424–973 ms |
| `/transactions` | 168–761 ms | 565–1036 ms |
| `/addresses` | 220–647 ms | 651–1025 ms |
| `/assets/policy` | 466 ms – 1.4 s | 493–889 ms |

So the multi-second numbers are specific to extreme units; normal traffic is sub-second, and the ordering fixes
added no latency.

---


extensions/blockfrost/asset/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/asset/storage/impl/BFAssetStorageReaderImpl.java
```diff
@@ -166,103 +162,144 @@ public Optional<BFAssetInfo> findAssetInfo(String unit) {
...
+                ? TRANSACTION.TX_INDEX.desc().nullsLast()
+                : TRANSACTION.TX_INDEX.asc().nullsFirst();
         SortField<?> txOrder = order == Order.desc ? ASSETS.TX_HASH.desc() : ASSETS.TX_HASH.asc();
 
         var query = dsl.select(ASSETS.TX_HASH, ASSETS.MINT_TYPE, ASSETS.QUANTITY)
                 .from(ASSETS)
+                .join(TRANSACTION).on(TRANSACTION.TX_HASH.eq(ASSETS.TX_HASH))
                 .where(assetCondition)
-                .orderBy(slotOrder, txOrder)
+                .orderBy(slotOrder, txIndexOrder, txOrder)
```
Sotatek-HuyLe3a: you can remove 'txOrder'

ducpm2303: Good catch, we dont need order by txHash anymore, removed in all method query

Sotatek-HuyLe3a:
@satran004 
## Follow-up implementation and validation after `cb2a591`

I continued the work on this PR after merging the latest `main` at `cb2a591`. The additional work focused on making the PostgreSQL queries maintainable, fixing deterministic pagination, verifying the assumed indexes, and validating correctness/performance against the live Blockfrost API.

### Additional commits

| Commit | Change |
|---|---|
| `00575f385` | Document the slot-window, page-boundary expansion, `tx_index` ordering, and jOOQ placeholder mapping used by the asset transaction pagination query. |
| `db3eeb072` | Document the result cardinality, ordering, and holder aggregation semantics of the asset history and address queries. |
| `7aebe96fb` | Document the policy and first-seen asset listing queries, including their pagination and ordering constraints. |
| `9fc52ab67` | Add the PostgreSQL partial index `idx_assets_mint_unit_slot (unit, slot) WHERE mint_type = 'MINT'` to `blockfrost-index.yml`. |
| `57220af1c` | Enforce deterministic ordering for global `/assets` by propagating and ordering the final result by the computed row number. |

The final `/assets` ordering change is necessary because an `ORDER BY` inside a derived table does not guarantee the order returned by the outer query. The returned page is now explicitly ordered by the same `(slot, tx_index, unit)` ranking used to select it.

### Index findings

Two indexes are important for the asset endpoints:

- `idx_assets_mint_unit_slot` was assumed to exist during the original policy-query optimization, but it was not included in the Blockfrost index configuration. It is now included as a PostgreSQL-only partial index.
- `idx_address_utxo_amounts` is already declared in `extra-index.yml` and is required by the unit-containment predicates used by `/addresses`, `/txs`, and `/transactions`.

Query-plan validation confirms that unit-containment predicates on `address_utxo.amounts` use a `Bitmap Index Scan` on `idx_address_utxo_amounts`.

The published performance of `/addresses`, `/txs`, and `/transactions` therefore depends on applying the GIN index documented in #958 as part of the Blockfrost extra-index setup.

### Latest benchmark results

All measurements below are Yaci Store response times with the required Blockfrost indexes available.

| Endpoint | Preprod | Preview, typical assets | Preview, high-cardinality asset |
|---|---:|---:|---:|
| `/assets/{asset}/txs` | 15–293 ms | 17–259 ms | 670–1,217 ms |
| `/assets/{asset}/transactions` | 18–91 ms | 18–129 ms | 619–840 ms |
| `/assets/{asset}/addresses` | 15–139 ms | 27–247 ms | 1.54–8.93 s |
| `/assets/policy/{policy_id}` | 15–119 ms | 13–153 ms | — |
| Global `/assets` | 763–1,664 ms | 396–813 ms | — |

The global `/assets` endpoint improved from approximately **8.94–9.82 seconds to 0.76–1.66 seconds on preprod**, and from approximately **11.36–11.95 seconds to 0.40–0.81 seconds on preview**, after the complete Blockfrost index set was available.

Mainnet was validated separately at the same chain tip as Blockfrost:

| Mainnet scenario | Yaci Store result |
|---|---:|
| Global `/assets`, pages 1–2, ascending and descending | 5.82–9.07 s |
| Policy containing 292,459 units | 10.15 s ascending; 1.47 s descending |
| High-cardinality asset `/transactions` | 6.16–8.70 s |
| High-cardinality asset `/txs`, ascending pages 1–2 | Exceeded the 30 s client timeout |
| High-cardinality asset `/addresses` | Exceeded the 30 s client timeout |

The partial MINT index was also measured independently from the HTTP comparison. For the 292,459-unit policy, the full query improved from **6.14–9.04 seconds to 2.27–2.34 seconds** and changed to a parallel index-only scan. The global first-mint query, which had previously exceeded the 30-second client timeout, completed in approximately **6.14 seconds** with the partial index.

Global `/assets` matched Blockfrost for all four combinations of pages 1–2 and ascending/descending order. The large policy also returned the same 100 units in the same positions for both directions. The latency variation between the controlled index benchmark and the later API run indicates that the mainnet policy query remains sensitive to ordering direction, cache state, and concurrent database load.

### Comparison with the original results

The new measurements broadly confirm the results reported by `ducpm2303`:

- The original preprod result of approximately **0.3 seconds** for asset transaction queries is consistent with the current **15–293 ms** range.
- The original preview result of approximately **1 second** is consistent with the current **670–1,217 ms** high-cardinality result. Normal preview assets are faster.
- Policy queries on preprod and preview are within the same range as the original **53–259 ms** measurements.
- The measured **6.16–8.70 seconds** for mainnet high-cardinality `/transactions` is consistent with the original mainnet worst-case result of approximately **6.2–6.4 seconds**.
- The mainnet timeouts and preview `/addresses` outliers confirm the original note that assets with many holders need a separate summary-table or aggregation optimization if consistently low latency is required.
- The partial-index benchmark confirms the original mainnet policy optimization, while the later **1.47–10.15 second** API range shows that worst-case stability still needs follow-up.

### Correctness comparison

Latest live Blockfrost comparisons:

- Mainnet asset suite: **80 OK / 4 DIFF / 0 ERR**, plus global `/assets`: **4 OK / 0 DIFF / 0 ERR**
- Preprod asset suite: **101 OK / 4 DIFF / 0 ERR**, plus global `/assets`: **4 OK / 0 DIFF / 0 ERR**
- Preview asset suite: **102 OK / 3 DIFF / 0 ERR**, plus global `/assets`: **4 OK / 0 DIFF / 0 ERR**

The remaining asset differences are known data/model differences rather than ordering regressions:

- Asset metadata fields are not currently populated by Yaci Store in the same way as Blockfrost.
- An empty asset name may be returned as `""` by Yaci Store and `null` by Blockfrost.
- The observed quantity differences were data consistency differences rather than ordering regressions; the returned unit ordering still matched Blockfrost.

### Verification performed

- Blockfrost asset module test task: successful
- Index configuration loader test: successful
- Live comparisons against Blockfrost on mainnet, preprod, and preview
- Ascending and descending pagination checked for the global asset endpoint
- Query plans checked for both the partial MINT index and the `address_utxo.amounts` GIN index

### Follow-up items

These are intentionally outside the current merge scope:

- Investigate a summary-table or equivalent optimization for `/assets/{asset}/addresses` on very high-cardinality assets.
- Add PostgreSQL integration coverage for the complex asset pagination queries.
- Revisit H2/MySQL query parity; PostgreSQL is the currently validated and prioritized path.
- Persist `tx_index` in the relevant tables so these queries no longer need to derive it through joins; see #1033.
- Add server-side statement cancellation/timeout so an HTTP timeout does not leave an expensive PostgreSQL query running.
- Ensure `extra-index.yml` is applied after sync/adapot calculation reaches tip. For large live databases, the GIN index should be created operationally with `CREATE INDEX CONCURRENTLY`.

From the PostgreSQL correctness and performance perspective, the implementation is now consistent with the original author's results once the same required indexes are present. The remaining items are follow-up improvements rather than blockers for the current PR.

satran004 (approved):
LGTM

PR #986, opened 2026-06-11
[Backport] Set restart policy for postgres and yaci-store docker containers

huylevt:
Check: https://github.com/bloxbean/yaci-store/pull/957


satran004 (approved):
Thanks @huylevt

PR #989, opened 2026-06-12
fix(governance): correct hard fork SPO abstain tally

huylevt:
original PR: #987


satran004 (approved):
Thanks @huylevt . LGTM

PR #998, opened 2026-06-24
fix: reset auto-restart counter on a sliding window 

### LINKED ISSUE #996: Indexer permanently stops retrying after `maxAttempts` and needs a manual restart
> All log evidence below is from a real mainnet run; every figure is reproducible with the included `grep` commands. Log timestamps are UTC.

## Summary
The cardano-node stop accidently and recover after more than 10mins. But yaci-indexer not recover automatically and I have to restart the docker image to fix it. Currently this problem show up two times.

`RequiredRestartProcessor` tracks auto-restart attempts with an in-process `AtomicInteger restartCount`. Once it reaches `store.auto-restart.max-attempts` (default 5), the indexer **stops retrying for good**: from then on it only logs `Max restart attempts reached. Manual intervention required.` every 20 seconds, until the **whole process is restarted by hand** — no matter how much time passes or whether the node has long since recovered.

For a feature whose whole point is *auto-recovery* (see #631 / #632), "give up forever after N tries" defeats the purpose: it implements what should be a **rate limit** (to prevent restart storms) as a **lifetime cap** ("N restarts for the entire life of the process").

## Environment

- yaci-store: deployed via the cardano-rosetta-java integration (exact version TBD; behavior verified against current `main`)
- Network: Cardano mainnet
- Deployment: Docker
- Relevant config: `store.auto-restart.max-attempts=5` (default)
- Sync protocol: Node-to-Node (N2N) Chain-Sync

## Reproduction & Evidence

> Logs are in Docker JSON format; the embedded timestamps are **UTC**.

### 1) The chain tip froze, and all 5 auto-restarts "succeeded" but achieved nothing

**For an unknown reason, the upstream Cardano node stopped returning new blocks** — the chain tip stayed at `slot 190252774` and never advanced (this issue isn't about *why* the node got stuck; it's about how the indexer reacts to it). The health check then triggered auto-restarts, but **each of the 5 restarts successfully connected, completed the handshake, and re-found the same frozen tip `slot 190252774`** — proving `start()` itself never failed. What "failed" was the tip not advancing, so restarting was the wrong medicine:

| # | Time (UTC) | Event |
|---|---|---|
| 1 | 2026-06-18T21:47:27Z | Intersect found at slot **190252774** |
| 2 | 2026-06-18T21:49:51Z | Intersect found at slot **190252774** |
| 3 | 2026-06-18T21:52:21Z | Intersect found at slot **190252774** |
| 4 | 2026-06-18T21:55:21Z | Intersect found at slot **190252774** |
| 5 | 2026-06-18T21:58:41Z | Intersect found at slot **190252774** |

Reproduce:
```bash
grep "Intersect found at slot: 190252774" yaci-indexer.log
# → appears only in the 5 clusters above; during normal sync the intersect slot keeps increasing
```

### 2) After the 5th attempt: the same line floods the log, frozen for ~80.8 hours

Once the 5th attempt is used up, the indexer is stuck for good: **it makes no further connection attempts** and just prints the same ERROR every **20 seconds**, until the process is restarted by hand 80.8 hours later. What an operator sees is this one line repeating forever:

```text
2026-06-18T22:00:56.809Z ERROR --- RequiredRestartProcessor : Max restart attempts reached. Manual intervention required.
2026-06-18T22:01:16.807Z ERROR --- RequiredRestartProcessor : Max restart attempts reached. Manual intervention required.
2026-06-18T22:01:36.807Z ERROR --- RequiredRestartProcessor : Max restart attempts reached. Manual intervention required.
... one every 20s, for 80.8 hours, 14544 lines total ...
2026-06-22T06:48:36.807Z ERROR --- RequiredRestartProcessor : Max restart attempts reached. Manual intervention required.
```

| Item | Value (UTC) |
|---|---|
| First line | 2026-06-18T22:00:56.809Z |
| Last line (just before the manual restart) | 2026-06-22T06:48:36.807Z |
| Count / interval | **14544** lines, one every **20 s** (= `store.admin.health-check-interval=20`) |
| Duration | ≈ **80.8 hours** |

Reproduce:
```bash
grep -c "Max restart attempts reached" yaci-indexer.log        # → 14544
grep "Max restart attempts reached" yaci-indexer.log | head -3 # consecutive lines are 20s apart
```

Throughout this window the health check fires a `RequiredSyncRestartEvent` every 20 s, but each one hits the retry-limit check and returns immediately — **not even a reconnect is attempted**. Auto-recovery is effectively dead.

## Root Cause

`components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java`

`restartCount` only ever increments. The **only** place it resets to zero is `startHealthMonitoring()`, and its reset condition is "5 continuous minutes of receiving new blocks". Once the count hits the cap, it gives up permanently:

```java
// retry-limit: give up permanently once the cap is reached
if (restartCount.get() >= storeProperties.getAutoRestartMaxAttempts()) {
    log.error("Max restart attempts reached. Manual intervention required.");
    return;
}
```

That reset condition is exactly where it breaks: **when the tip is frozen there are no new blocks**, so "5 continuous minutes of blocks" can never be satisfied and `restartCount` is never reset. Worse, that monitoring thread also has a hard 10-minute timeout — once an outage lasts longer than 10 minutes the thread exits on its own, so it isn't even waiting to reset anymore. Together, `restartCount` is stuck at the cap and auto-recovery is dead.

> Corroborating evidence: two brief reconnects earlier that day (06-18 06:40 and 10:38 UTC) were each followed by blocks resuming and the counter being reset successfully; only the 21:47 incident exhausted the counter, because the frozen tip never produced the "healthy" evidence the reset needs. This confirms the mechanism works only while blocks are flowing, and fails in exactly the long-outage case it is needed for.

## Proposed Fix: a sliding window

First, an important premise: **the yaci-store process does not exit when `maxAttempts` is hit** — it just stops reconnecting and idles, flooding the log. Since the process stays alive, "wait a while and try reconnecting a few more times" costs almost nothing — a reconnect is just a disconnect, a handshake, and an intersection lookup, which is cheap. So the right behavior is not to *give up*, but to *back off and try again later*.

With that in mind, change `restartCount` from a *lifetime cap* into a **rate limit within a sliding window**: if more than some window (e.g. 10 minutes) has passed since the last restart, treat the new outage as a fresh incident and reset the counter.

The core change extracts the decision into a method `evaluateRestartAttempt(now, event)` (with `now` as a parameter, so it is easy to unit test), inserting a "sliding-window reset" between the existing debounce and retry-limit checks. It **reuses the existing `lastRestartAttempt` field and adds no new state**:

```java
static final long RESTART_WINDOW_MS = 600000; // 10-minute sliding window

int evaluateRestartAttempt(long now, RequiredSyncRestartEvent event) {
    long lastAttempt = lastRestartAttempt.get();

    // (1) debounce (existing): ignore events that arrive too close together
    if (now - lastAttempt < storeProperties.getAutoRestartDebounceWindowMs())
        return 0;

    // (2) new: more than a window since the last restart -> reset the counter (fresh incident)
    if (lastAttempt > 0 && now - lastAttempt > RESTART_WINDOW_MS)
        restartCount.getAndSet(0);

    // (3) retry-limit (existing): pause if over the cap within the window
    if (restartCount.get() >= storeProperties.getAutoRestartMaxAttempts())
        return 0;

    lastRestartAttempt.set(now);
    return restartCount.incrementAndGet(); // the attempt number
}
```

Behavior comparison:

| | Current (bug) | After the fix |
|---|---|---|
| After `maxAttempts` is hit | Errors every 20 s forever; needs a manual restart | Pauses for ~10 minutes |
| Once the window passes | Still stuck | Counter resets → retries `maxAttempts` more times |
| Long outage | Gives up forever | One burst of up to `maxAttempts` every ~10 minutes, self-healing indefinitely |

The "restart-storm protection" intent of `maxAttempts` is preserved (it still caps the restart rate within a window); it is just no longer "N times for the lifetime of the process".

This fix also removes `startHealthMonitoring()`: the sliding window takes over the "reset the counter" job, so keeping it would just leave two overlapping 10-minute mechanisms. As a result the reset condition changes from *"5 minutes of confirmed healthy sync"* to *"more than a window since the last restart"* — the main effect being that the indexer **no longer gives up permanently**, which is the desired behavior for an unattended service.

If keeping the health-confirmation behavior is preferred, an alternative is "sliding window + a one-shot health probe after a successful restart" (at the cost of reintroducing the `HealthService` dependency). And `RESTART_WINDOW_MS` is currently a hard-coded constant (to keep new config to a minimum); if it should be configurable, a field can be added to `YaciStoreProperties.AutoRestart` alongside the existing `maxAttempts` / `backoffBaseMs`.

## Related

- #660 *Indexer stopped in between on mainnet* — same symptom (`Max restart attempts reached`); likely the same root cause via a different trigger.
- #959 *Auto-recovery deadlock: AutoRecoveryStartService bails on scheduleToStop=true* — a related defect in the same auto-recovery module, with a different root cause.

## Implementation & PR

The full change (fix + unit tests) is implemented and can be viewed on this branch:

**https://github.com/DanchuoZhong/yaci-store/commits/fix/auto-restart-sliding-window**

Happy to adjust based on feedback on the design trade-offs above before opening the formal PR.

DanchuoZhong:
## Description

`RequiredRestartProcessor` uses an in-memory counter (`restartCount`) that acts as a 
**lifetime cap**: once exhausted at `max-attempts` (default 5), the indexer stops 
retrying forever — even after the outage clears. In one incident, this left the 
indexer stuck for 80+ hours until manual restart.

This PR converts it to a **sliding-window rate limit**: the counter resets if no 
restart attempt occurs within 10 minutes, so each outage starts fresh. `max-attempts` 
becomes a per-window limit, preserving restart-storm protection.

Closes #996 
### Changes
- Extract gate logic into `evaluateRestartAttempt()` for testability
- Add sliding-window reset: if idle > 10min, reset counter and allow fresh attempts
- Remove now-redundant `startHealthMonitoring()` / `HealthService`
- Downgrade exhausted-attempts log: `ERROR` → `WARN` (automatic recovery expected)
- Add `RequiredRestartProcessorTest` (5 deterministic cases, no real waiting)
- Update docs: `max-attempts` now described as per-window, not lifetime

### Notes for reviewers
- Log text changed from "Manual intervention required" to "Pausing until window resets"
- `RESTART_WINDOW_MS` (10 min) is hard-coded; can be made configurable if needed
- Closes #<issue-number>

## Type of Change
- [x] Bug fix

## Testing
- [x] Full suite: `./gradlew test` → BUILD SUCCESSFUL, 47 modules
- [x] New tests: 5 cases in `RequiredRestartProcessorTest` (debounce, rate-limit, window reset)

## Checklist
- [x] Code follows project standards
- [x] Documentation updated
- [x] No new warnings

### DIFF
--- a/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
+++ b/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
@@ -27,10 +27,6 @@ public class RequiredRestartProcessor {
     private final AtomicInteger restartCount = new AtomicInteger(0);
     private final Lock restartLock = new ReentrantLock();
 
-    // If there's been no restart for this long, the attempt counter is reset so a new
-    // outage starts fresh instead of inheriting the count of an old, resolved one.
-    static final long RESTART_WINDOW_MS = 600000; // 10 minutes
-
     @EventListener
     public void handleRequiredRestart(RequiredSyncRestartEvent event) {
         Thread.startVirtualThread(() -> handleRestartInterval(event));
@@ -78,26 +74,27 @@ public class RequiredRestartProcessor {
      * without waiting.
      */
     int evaluateRestartAttempt(long now, RequiredSyncRestartEvent event) {
-        long lastAttempt = lastRestartAttempt.get();
+        long lastAttemptEpochMs = lastRestartAttempt.get();
+        long windowMs = storeProperties.getAutoRestartWindowMs();
 
-        if (now - lastAttempt < storeProperties.getAutoRestartDebounceWindowMs()) {
+        if (now - lastAttemptEpochMs < storeProperties.getAutoRestartDebounceWindowMs()) {
             log.info("Within debounce window. Ignoring restart event: {}", event.getReason());
             return 0;
         }
 
         // A gap longer than the window means the previous burst is over, so start counting again.
-        if (lastAttempt > 0 && now - lastAttempt > RESTART_WINDOW_MS) {
+        if (lastAttemptEpochMs > 0 && now - lastAttemptEpochMs > windowMs) {
             int previous = restartCount.getAndSet(0);
             if (previous > 0) {
                 log.info("No restart attempt in the last {} ms. Resetting restart attempt counter from {} to 0.",
-                         RESTART_WINDOW_MS, previous);
+                         windowMs, previous);
             }
         }
 
         if (restartCount.get() >= storeProperties.getAutoRestartMaxAttempts()) {
             log.warn("Reached max restart attempts ({}) within the last {} ms. " +
                      "Pausing further restarts until the window resets.",
-                     storeProperties.getAutoRestartMaxAttempts(), RESTART_WINDOW_MS);
+                     storeProperties.getAutoRestartMaxAttempts(), windowMs);
             return 0;
         }
 
--- a/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
+++ b/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
@@ -2,7 +2,6 @@ package com.bloxbean.cardano.yaci.store.core.processor;
 
 import com.bloxbean.cardano.yaci.store.common.config.StoreProperties;
 import com.bloxbean.cardano.yaci.store.core.annotation.ReadOnly;
-import com.bloxbean.cardano.yaci.store.core.service.HealthService;
 import com.bloxbean.cardano.yaci.store.core.service.StartService;
 import com.bloxbean.cardano.yaci.store.events.internal.RequiredSyncRestartEvent;
 import lombok.RequiredArgsConstructor;
@@ -11,7 +10,6 @@ import org.springframework.context.event.EventListener;
 import org.springframework.stereotype.Component;
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
@@ -23,21 +21,18 @@ import java.util.concurrent.locks.ReentrantLock;
 @Slf4j
 public class RequiredRestartProcessor {
     private final StartService startService;
-    private final HealthService healthService;
     private final StoreProperties storeProperties;
 
     private final AtomicLong lastRestartAttempt = new AtomicLong(0);
     private final AtomicInteger restartCount = new AtomicInteger(0);
     private final Lock restartLock = new ReentrantLock();
-    private final AtomicBoolean monitoringHealth = new AtomicBoolean(false);
 
-    // Track successful sync period
-    private static final long SUCCESS_THRESHOLD_MS = 300000; // 5 minutes
-    private static final long HEALTH_CHECK_INTERVAL_MS = 30000; // 30 seconds
+    // If there's been no restart for this long, the attempt counter is reset so a new
+    // outage starts fresh instead of inheriting the count of an old, resolved one.
+    static final long RESTART_WINDOW_MS = 600000; // 10 minutes
 
     @EventListener
     public void handleRequiredRestart(RequiredSyncRestartEvent event) {
-
         Thread.startVirtualThread(() -> handleRestartInterval(event));
     }
 
@@ -53,22 +48,11 @@ public class RequiredRestartProcessor {
         }
 
         try {
-            // Check debounce window
-            long now = System.currentTimeMillis();
-            long lastAttempt = lastRestartAttempt.get();
-            if (now - lastAttempt < storeProperties.getAutoRestartDebounceWindowMs()) {
-                log.info("Within debounce window. Ignoring restart event: {}", event.getReason());
-                return;
+            int attemptNumber = evaluateRestartAttempt(System.currentTimeMillis(), event);
+            if (attemptNumber <= 0) {
+                return; // debounced or rate-limited
             }
 
-            // Check retry limit
-            if (restartCount.get() >= storeProperties.getAutoRestartMaxAttempts()) {
-                log.error("Max restart attempts reached. Manual intervention required.");
-                return;
-            }
-
-            // Calculate backoff
-            int attemptNumber = restartCount.incrementAndGet();
             long backoffMs = calculateBackoff(attemptNumber);
 
             log.info("Scheduling sync restart. Reason: {}, Attempt: {}, Backoff: {}ms",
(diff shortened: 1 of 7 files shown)


components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
```diff
@@ -80,15 +64,47 @@ private void handleRestartInterval(RequiredSyncRestartEvent event) {
...
     }
 
+    /**
+     * Applies the debounce window, the sliding-window counter reset and the
+     * max-attempts limit. Returns the 1-based attempt number when a restart should go
+     * ahead, or 0 to skip. {@code now} is a parameter so the windowing can be tested
+     * without waiting.
+     */
+    int evaluateRestartAttempt(long now, RequiredSyncRestartEvent event) {
+        long lastAttempt = lastRestartAttempt.get();
```
matiwinnetou: cosmetics: its not clear immediately from reading the code that lastAttempt is in fact unix time, maybe variable name change? I know it was named like this before though... but this code is also new to me.

components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
```diff
@@ -80,15 +64,47 @@ private void handleRestartInterval(RequiredSyncRestartEvent event) {
...
+
+        if (now - lastAttempt < storeProperties.getAutoRestartDebounceWindowMs()) {
+            log.info("Within debounce window. Ignoring restart event: {}", event.getReason());
+            return 0;
+        }
+
+        // A gap longer than the window means the previous burst is over, so start counting again.
+        if (lastAttempt > 0 && now - lastAttempt > RESTART_WINDOW_MS) {
```
matiwinnetou: Why in some cases you are using a static variable and in some cases configurable variable, is there any reason why value is static and the other is configurable?

RESTART_WINDOW_MS: static
getAutoRestartDebounceWindowMs: configurable

matiwinnetou:
Your contribution and detailed descriptions, analysis and fix are highly appreciated.

LGTM (Looks Good To Me), however we must not forget it also requires real world testing, not only unit test checks.

components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
DanchuoZhong: @matiwinnetou  Yes the name not changed because it named like this before. Indeed renanmed it will make it clear.

components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/processor/RequiredRestartProcessor.java
DanchuoZhong: @matiwinnetou No strong reason — I initially kept it as a constant to avoid adding config, but I agree it's better to make it configurable. I'll promote it to store.auto-restart.window-ms

DanchuoZhong:
> Your contribution and detailed descriptions, analysis and fix are highly appreciated.
> 
> LGTM (Looks Good To Me), however we must not forget it also requires real world testing, not only unit test checks.

@matiwinnetou 
Thanks! Out of curiosity — what kind of **real-world testing** did you have in mind here? A docker-compose run of cardano-rosetta-java with new branch of yaci-store ,and manually stop caradano-node for like 10mins and start it again  to see if yaci-indexer will continue working ? Just want to make sure I'm testing the right scenario.

huylevt:
Nice work! This nails the root cause from #996. LGTM @DanchuoZhong !

huylevt (approved):

satran004:
Thanks @DanchuoZhong

satran004:
@huylevt Let's plan to merge it back to release/2.0.x branch also

PR #999, opened 2026-06-25
docs: simplify scope tutorial queries

huylevt:
## Summary

  - Fix the Governance Watch vote tally SQL by adding a `total_votes` alias and ordering by it instead of using aggregate aliases inside an expression.
  - Simplify the NFT mint tracking tutorial by removing the advanced metadata policy-filter callout.
  - Remove redundant query filters from the NFT tutorial where the plugin filters already scope stored rows.


satran004 (approved):
LGTM

PR #1002, opened 2026-07-04
feat(transaction): store full transaction CBOR behind save-full-tx-cbor

edridudi:
## Summary

Adds a `store.transaction.save-full-tx-cbor` option. Today `save-cbor` stores the transaction **body** CBOR only (`transaction.getBody().getCbor()`). Consumers that need to render/verify redeemers, Plutus scripts, datums, or metadata require the **full signed transaction** CBOR (`body + witness_set + is_valid + auxiliary_data`). This flag makes the transaction CBOR store persist that full tx, reassembled from the already-parsed transaction — no extra node round-trips.

## Design

- New `FullTxCborReassembler` builds the transaction CBOR **envelope at the `co.nstant.in.cbor` DataItem level**: `[ body, witness_set, is_valid, auxiliary_data ]`.
  - The **body is kept byte-for-byte verbatim** — the original `getBody().getCbor()` bytes are decoded into a `DataItem` and placed into the array as-is, never re-serialized through the cardano-client-lib `TransactionBody` model. This avoids reshaping/dropping any body field the CCL model may not (yet) cover.
  - The witness set is built from the parsed `Witnesses` via cardano-client-lib: vkey + **bootstrap** witnesses, native + Plutus V1/V2/V3 scripts, redeemers, and witness datums. Auxiliary data carries metadata **and** its native/Plutus scripts.
- **Best-effort, non-fatal:** a single malformed/unmappable component is skipped (debug-logged), never aborting reassembly. Any hard failure falls back to storing body-only CBOR.
- `save-full-tx-cbor` gates the CBOR collect/save path independently of `save-cbor`.

## Backward compatibility

- Flag **off** (default): behaviour is unchanged.
- Historical rows written under `save-cbor` remain body-only; flipping the flag changes stored bytes for **new** rows only — operators should treat `transaction_cbor.cbor` as body-only-or-full and version/migrate accordingly.

## Tests

`FullTxCborReassemblerTest` — verbatim-body byte equality, vkey/bootstrap witnesses, native + Plutus scripts, redeemers, witness datums, auxiliary-data metadata + native + Plutus scripts, and full `Transaction.deserialize` round-trip; plus processor gating tests. `./gradlew :stores:transaction:test` passes.

## Notes

- Reassembled CBOR is intended for display/verification of tx contents; vkey signatures and canonical byte ordering of the *witness set* are not guaranteed to match the on-chain bytes (the **body** is verbatim). Happy to adjust scope/naming to match project conventions.
- Field-mapped against yaci-core model getters and cardano-client-lib `transaction.spec`.

Signed-off-by included.

### DIFF
--- a/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionAutoConfigProperties.java
+++ b/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionAutoConfigProperties.java
@@ -46,6 +46,17 @@ public class TransactionAutoConfigProperties {
       private boolean saveCbor = false;
 
       /**
+       * Enable/disable saving of the full signed transaction CBOR (body + witness set + isValid + auxiliary data)
+       * instead of the transaction body-only CBOR.
+       * When enabled, the indexer reassembles the full transaction CBOR (CBOR array, tag 84) from its
+       * constituent parts and stores that instead of the body-only CBOR (CBOR map, tag a4).
+       * This is useful for wallets and clients that need the complete signed transaction, e.g. for re-broadcast
+       * or witness verification.
+       * Note: Has no effect unless {@code saveCbor} is also enabled.
+       */
+      private boolean saveFullTxCbor = false;
+
+      /**
        * Enable/disable pruning of block CBOR data.
        * When enabled, CBOR data older than cborPruningSafeSlots will be automatically deleted.
        */
--- a/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionStoreAutoConfiguration.java
+++ b/starters/transaction-spring-boot-starter/src/main/java/com/bloxbean/cardano/yaci/store/starter/transaction/TransactionStoreAutoConfiguration.java
@@ -28,6 +28,7 @@ public class TransactionStoreAutoConfiguration {
         transactionStoreProperties.setPruningSafeSlots(properties.getTransaction().getPruningSafeSlots());
         transactionStoreProperties.setSaveWitness(properties.getTransaction().isSaveWitness());
         transactionStoreProperties.setSaveCbor(properties.getTransaction().isSaveCbor());
+        transactionStoreProperties.setSaveFullTxCbor(properties.getTransaction().isSaveFullTxCbor());
         transactionStoreProperties.setCborPruningEnabled(properties.getTransaction().isCborPruningEnabled());
         transactionStoreProperties.setCborPruningSafeSlots(properties.getTransaction().getCborPruningSafeSlots());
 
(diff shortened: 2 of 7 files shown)


edridudi:
Addressed the SonarCloud findings (force-pushed):

- **Cognitive Complexity (CRITICAL ×3)** — `FullTxCborReassembler.buildWitnessSet` and its helpers were refactored: each witness-set component (vkey, bootstrap, native/Plutus V1-3 scripts, redeemers, datums) is now its own small private mapper, and `TransactionProcessor` extracts `collectCborIfEnabled`/`saveCollectedCbor`. Behavior-preserving (per-component try/catch-skip + body-only fallback unchanged); `:stores:transaction:test` stays green.
- **`FullTxCborReassemblerTest:26` (MAJOR, "commented-out code")** — this is a false positive: line 26 is a plain-English explanatory comment about the test fixtures (structurally-valid-but-fake vkey/signature hex), not commented-out code. Left as-is since it's useful documentation; happy to mark it won't-fix or reword if you prefer.

Kept the diff minimal to the feature — no unrelated cleanups.

satran004:
Thanks @edridudi for this PR.  I think this will be a useful feature, and the PR is well put together: the flag/config wiring, processor and the fallback behavior are all good to merge as-is.
  
  The one part I'd like to change is where the transaction gets reassembled. Rebuilding the witness set and auxiliary data from parsed model objects through cardano-client-lib can't guarantee byte equality with the on-chain transaction — witness map ordering and encoding choices, native scripts round-tripping through JSON, and auxiliary data losing its original on-chain shape (plain metadata map vs `[metadata, scripts]` vs tag-259 map) can all produce bytes that differ from what was submitted, even when they decode to an equivalent transaction. We hit this same class of issue in yaci itself with datum/redeemer hashes (bloxbean/yaci#37), and the fix there was to slice the exact original bytes from the block rather than re-serialize.

  Since only yaci holds the original block bytes, I have created this PR to handle the assembly in yaci: bloxbean/yaci#169 adds a `YaciConfig.returnFullTxCbor` flag that splices the exact body, witness-set and auxiliary-data segment bytes from the block per the era CDDL, and exposes the result as `Transaction.getTxCbor()`. No CBOR encoder touches the on-chain bytes, so the output is byte-exact by construction, and it fails closed (`null`) rather than ever producing wrong CBOR.

  Once a full sync test against all networks passes, we'll do a quick yaci release. This PR can then be simplified to keep the `save-full-tx-cbor` configuration and processor changes, enable the yaci flag. `FullTxCborReassembler` won't be needed anymore.

edridudi:
so @satran004 can we close this PR then ?

satran004:
@edridudi I completed a full mainnet sync with the Yaci changes, and everything looks good. We plan to merge the Yaci PR soon once @huylevt finishes the review: https://github.com/bloxbean/yaci/pull/169

Meanwhile, if you can build Yaci locally and update this PR, that would be great. We don't need the assembly changes in this PR—only the config/autoconfig changes for the new property.

satran004:
@edridudi Yaci 0.4.5 is now available https://github.com/bloxbean/yaci/releases/tag/v0.4.5 . We can use that in this PR.

satran004:
@Sotatek-HenryDo Can you look into this PR? We need to remove assembly part from this PR as it is now handled in Yaci.

Issue #1013, opened 2026-07-14
Committee accepted-ratio comparison rounds before threshold comparison, can diverge from cardano-ledger


huylevt:
## Summary

`CommitteeVotingEvaluator` computes the committee accepted ratio (`yes / (yes + no)`) as a `BigDecimal` rounded to **4 decimal places** (`RoundingMode.HALF_UP`) and then compares it against the committee threshold. `cardano-ledger` compares the same ratio as an **exact rational number** (`Ratio Integer`), never rounded. Near a threshold boundary, the rounding step can flip the pass/fail outcome relative to the ledger, causing yaci-store to record a governance proposal as `RATIFIED` when cardano-node/cardano-ledger would reject it (or vice versa).

## Example

- committee size: 3, committee YES votes: 2, threshold: `6667/10000`
- `2/3 = 0.666666...` vs `6667/10000 = 0.6667` -> ledger: `2/3 < 6667/10000` -> not accepted
- yaci-store: `2 / 3` rounded to 4 places (`HALF_UP`) -> `0.6667` -> compares equal to threshold -> incorrectly passes

PR #1014, opened 2026-07-15
fix(governance): compare committee ratios exactly

huylevt:
#1013
## Summary

  - Replace committee accepted-ratio decimal rounding with exact rational comparison.
  - Compare `yesVotes / totalExcludingAbstain` against the committee threshold by cross-multiplying `BigInteger` values.
  - Add a regression test for the `2/3` vs `6667/10000` threshold boundary that previously rounded up and passed incorrectly.

  ## Why

  `cardano-ledger` compares committee voting ratios as exact rationals. The previous implementation rounded the accepted ratio to 4 decimal places before comparison, which could diverge from ledger behavior near
  threshold boundaries.

  ## Tests

  - `./gradlew :aggregates:governance-rules:test --tests '*CommitteeVotingEvaluatorTest.evaluate_usesExactRationalComparisonAtThresholdBoundary'`
  - `./gradlew :aggregates:governance-rules:test --tests '*CommitteeVotingEvaluatorTest'`
  - `./gradlew :aggregates:governance-rules:test`


satran004 (approved):
LGTM.

Issue #1017, opened 2026-07-17
[Governance] SPO AlwaysNoConfidence non-voters counted as No/Yes instead of Abstain in bootstrap phase


huylevt:
## Summary

While checking Conway SPO tally behavior against `cardano-ledger`,  a mismatch is found in `VoteTallyCalculator.computeSPOTallies` for non-voting SPO stake during the Conway bootstrap phase.

The ledger rule in `eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs` (`spoAcceptedRatio`) (cardano-ledge) applies non-voter guards in this order:

```haskell
case vote of
  Nothing
    | HardForkInitiation {} <- pProcGovAction -> (yes, abstain)
    | hardforkConwayBootstrapPhase pv -> (yes, abstain + stake)
    | otherwise -> case defaultStakePoolVote poolId reStakePools reAccounts of
        DefaultNoConfidence
          | NoConfidence {} <- pProcGovAction -> (yes + stake, abstain)
        DefaultAbstain -> (yes, abstain + stake)
        _ -> (yes, abstain)
  Just Abstain -> (yes, abstain + stake)
  Just VoteNo  -> (yes, abstain)
  Just VoteYes -> (yes + stake, abstain)
```

So the guard order is:

1. `HardForkInitiation`: non-voters remain in the denominator.
2. Conway bootstrap phase: non-voters count as `Abstain` for every non-HFI action.
3. Post-bootstrap only: consult delegation defaults (`AlwaysNoConfidence`, `AlwaysAbstain`, regular/default no).

During bootstrap, a non-voting pool should count as `Abstain` for any non-HFI action regardless of its reward account's DRep delegation.

In yaci-store, `delegateToNoConfidenceDRep` is handled before the bootstrap rule:

- For `NO_CONFIDENCE`, it is added to `totalYesStake`.
- For other non-HFI actions, it is not added to abstain, so it becomes implicit `No`.
- Bootstrap handling only adds `doNotVote` to abstain, not `delegateToNoConfidenceDRep`.

Result:

- Bootstrap + non-HFI + AlwaysNoConfidence-delegated non-voter:
  - expected: `Abstain`
  - actual: implicit `No`
- Bootstrap + `NO_CONFIDENCE` + AlwaysNoConfidence-delegated non-voter:
  - expected: `Abstain`
  - actual: `Yes`

## Impact

[...]

PR #1020, opened 2026-07-19
[Governance rule] Fix SPO tally bootstrap default delegation

huylevt:
#1017 
 ## Summary

  Fix SPO vote tally behavior during the Conway bootstrap phase to match `cardano-ledger` non-voter guard ordering.

  During bootstrap, non-voting SPO stake delegated through default DRep options should be counted as abstain for non-HardForkInitiation actions, including stake delegated to the AlwaysNoConfidence DRep. Previously,
  AlwaysNoConfidence-delegated non-voting SPO stake was applied before the bootstrap rule, causing it to be counted as YES for `NO_CONFIDENCE` actions or implicit NO for other non-HFI actions.

  ## Changes

  - Updated `VoteTallyCalculator.computeSPOTallies` to apply SPO non-voter handling in ledger-compatible order:
    - HardForkInitiation: leave non-voting/default SPO stake as non-abstain/non-yes stake.
    - Bootstrap phase non-HFI actions: count `doNotVote`, AlwaysAbstain delegation, and AlwaysNoConfidence delegation as abstain.
    - Post-bootstrap: keep existing delegation default behavior.
  - Added regression coverage for:
    - BootstrapPhase non-HFI + AlwaysNoConfidence delegation -> abstain
    - BootstrapPhase `NO_CONFIDENCE` + AlwaysNoConfidence delegation -> abstain
    - BootstrapPhase HFI behavior
    - PostBootstrapPhase behavior for `NO_CONFIDENCE` and non-HFI actions

  ## Verification

  ```bash
  ./gradlew :aggregates:governance-rules:test

### DIFF
--- a/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/voting/VoteTallyCalculator.java
+++ b/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/voting/VoteTallyCalculator.java
@@ -75,18 +75,20 @@ public final class VoteTallyCalculator {
         BigInteger doNotVote = nz(spo.getDoNotVoteStake());
 
         BigInteger totalYesStake = yesVote;
-        if (type == GovActionType.NO_CONFIDENCE) {
-            totalYesStake = totalYesStake.add(delegateToNoConfidenceDRep);
-        }
-
         BigInteger totalAbstainStake = abstainVote;
-        if (type != GovActionType.HARD_FORK_INITIATION_ACTION) {
-            totalAbstainStake = totalAbstainStake.add(delegateAutoAbstainDRep);
-        }
 
-        // In bootstrap phase, all do not vote stake is considered as abstain stake except for HardForkInitiationAction
-        if (isInBootstrapPhase && type != GovActionType.HARD_FORK_INITIATION_ACTION) {
-            totalAbstainStake = totalAbstainStake.add(doNotVote);
+        if (type != GovActionType.HARD_FORK_INITIATION_ACTION) {
+            if (isInBootstrapPhase) {
+                totalAbstainStake = totalAbstainStake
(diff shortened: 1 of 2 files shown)


satran004 (approved):
LGTM

Sotatek-HuyLe3a:
Claude Code Suggestions

  1. Restore post-bootstrap HFI coverage. The old computeSPOTalliesHardForkInitTreatsNonVotingSPOsAsNoEvenWhenDelegatedToAlwaysAbstain test used isInBootstrapPhase=false; the PR flipped it to
  true, so the post-bootstrap HFI combination now has no dedicated test. The code path is currently phase-independent for HFI, but the ledger docs explicitly call out "HFI behaves the same in
  both phases" — a cheap regression test pins that against future refactors.
  2. Re-add an explanatory comment. The old code's comment was deleted and not replaced. This logic was subtle enough to be wrong once; a short comment stating the guard order and referencing
  spoAcceptedRatio in cardano-ledger's Ratify.hs would help the next reader (the DRep tally method above it has exactly this style of comment).
  3. Flag the data impact. This changes computed tallies for bootstrap-era proposals (mainnet protocol v9, pre-Plomin). If governance-aggr persists proposal stats/ratification results computed
  with the old logic (VotingStatsService / ProposalCollectionService feed stored data), those remain stale until re-processed — worth a release-note line or a reprocessing consideration for
  affected epochs.

  Out-of-scope observation (pre-existing, possible follow-up)

  SPOVotingDataCollector.buildEpochAggregates derives the "non-voting pools" set from votes across all proposals in scope, shared per epoch. A pool that voted on proposal A but not proposal B
  lands in B's doNotVote bucket rather than its delegation bucket. During bootstrap that's harmless (both → abstain), but post-bootstrap a such pool with an AlwaysAbstain delegation would be
  counted as No instead of abstain for B. Per-ledger, the default-vote lookup is per-action. Not introduced by this PR — just worth a follow-up issue.

PR #1021, opened 2026-07-20
feat: expose network identity and sync status

### LINKED ISSUE #1018: Expose network identity (protocol magic) via the API
## Problem

There is currently no way for an API client to determine **which Cardano network a yaci-store instance is serving**.

A client that can be pointed at different networks (mainnet / preprod / preview / a local devnet) has no way to verify that the instance it is talking to is the network it expects. Today this can only be assumed from configuration or user input, so a mistyped or stale URL silently produces wrong results instead of a clear error.

## What's available today

I checked the full API surface (119 paths in `/v3/api-docs`):

- `/api/v1/network` returns supply and stake totals only (Blockfrost's `/network` shape) — no magic.
- There is no `/genesis` endpoint.
- No other endpoint carries `network_magic`, a system start, or any network identifier.

The closest available signal is *inferring* it from `/api/v1/epochs/latest`:

```
epoch_length        = end_time - start_time
implied_system_start = start_time - epoch * epoch_length
```

and comparing against known genesis constants. That does work, but it is a heuristic built on genesis arithmetic rather than a fact the store reports.

## Proposal

Expose the protocol magic. Either:

**1. A Blockfrost-compatible `/api/v1/genesis`** (preferred — standard shape, already understood by existing BF clients):

```json
{
  "network_magic": 42,
  "system_start": 1784292534,
  "epoch_length": 600,
  "slot_length": 1,
  "security_param": 108
}
```

**2. Or add `network_magic` (and ideally `system_start`) to the existing `/api/v1/network`.**

Only `network_magic` is essential — clients can default or ignore the rest.

This is configuration yaci-store already holds (it needs the magic to talk to the node), so it should mostly be a matter of surfacing it.

## Why `/genesis` specifically

Blockfrost exposes `/genesis` with exactly these fields, so Blockfrost-compatible clients — e.g. Cardano Client Lib's Blockfrost backend, which Yaci DevKit already documents as a supported integration path — would get network verification for free, with no client-side special-casing.

It would also remove the need for clients to hardcode assumptions like "a devnet is always magic 42", which breaks as soon as someone runs a devnet on a different magic.

---

### Secondary (happy to split into a separate issue)

There is also no way to tell whether the store is **caught up with the node's tip**. `/api/v1/blocks/latest` gives the store's tip, but not the node's, so a client cannot distinguish "synced" from "still indexing" and can only report a block height. An indexer-lag indicator (store tip vs node tip, or a boolean/`lag_blocks`) would let clients show a truthful sync state.

satran004:
## Summary

- add a Blockfrost-compatible `GET /api/v1/genesis` endpoint exposing protocol magic and genesis parameters
- add `GET /api/v1/sync/status` with store-vs-node tip details and explicit handling when the node tip is unavailable
- refresh stale cached tips, back off failed node-tip requests, and avoid false `synced`/zero-lag results
- keep `slot_length` integer-compatible with the Blockfrost schema, including fractional devnet configurations
- wire the new core API module into the Spring Boot starter and update the admin UI for unavailable-tip states

## Verification

- `./gradlew :components:common:test :components:core:test :components:core-api:test :components:admin-ui:buildFrontend --rerun-tasks`
- `./gradlew :applications:all:bootJar`
- `git diff --check`

Closes #1018

### DIFF
--- a/components/admin-ui/frontend/src/lib/api/client.ts
+++ b/components/admin-ui/frontend/src/lib/api/client.ts
@@ -9,6 +9,7 @@ export interface SyncStatus {
     syncPercentage: number;
     networkBlock: number;
     networkSlot: number;
+    networkTipAvailable: boolean;
     synced: boolean;
     protocolMagic: number;
 }
--- a/components/admin-ui/frontend/src/lib/components/dashboard/SyncProgress.svelte
+++ b/components/admin-ui/frontend/src/lib/components/dashboard/SyncProgress.svelte
@@ -3,6 +3,7 @@
     export let isSynced: boolean;
     export let currentBlock: number;
     export let networkBlock: number;
+    export let networkTipAvailable: boolean;
 
     $: progressPercentage = Math.min(100, Math.max(0, percentage));
 </script>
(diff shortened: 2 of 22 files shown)


huylevt (approved):

huylevt:
LGTM! @satran004

PR #1024, opened 2026-07-21
Fix committee member expiry boundary

### LINKED ISSUE #1022: [Governance] Committee member expiry boundary should be inclusive
## Summary

Yaci Store appears to treat a committee member's `expiredEpoch` as an exclusive boundary when resolving active committee members. Cardano Ledger treats committee membership as valid through the expiry epoch itself. This can make Yaci Store drop a valid committee member one epoch too early, affecting committee vote tallying and committee minimum-size checks.

## Cardano Ledger Rule

The committee voting ratio treats a member as expired only when `currentEpoch > expiry`:

```haskell
accumVotes (!yes, !tot) member expiry
  | currentEpoch > expiry = (yes, tot) -- member is expired, vote "abstain" (don't count it)
```

Source: `cardano-ledger/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs:151-152`

The committee min-size check uses the same inclusive validity rule:

```haskell
isActive coldKey validUntil =
  case Map.lookup coldKey hotKeys of
    Just (CommitteeMemberResigned _) -> False
    Just _ -> currentEpoch <= validUntil
    Nothing -> False
```

Source: `cardano-ledger/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs:472-476`

Thus, a committee member is active through its expiry epoch.

## Yaci Store Behavior

Yaci Store filters active committee members with a strict greater-than check:

```java
.and(COMMITTEE_MEMBER.EXPIRED_EPOCH.gt(epoch)) // todo: verify 'greater than' or 'greater or equals'
```

Source: `yaci-store/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/storage/impl/CommitteeMemberStorageImpl.java:172`

The `expiredEpoch` is populated directly from the committee term:

```java
updateCommittee.getNewMembersAndTerms().forEach((credential, term) ->
        updatedCommitteeMembers.add(CommitteeMember.builder()
                .hash(credential.getHash())
                .startEpoch(epoch)
                .expiredEpoch(term)
```

Source: `yaci-store/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/CommitteeMemberProcessor.java:161-165`

If `expiredEpoch` corresponds to the Ledger `validUntil` epoch, the query excludes the member one epoch too early.

## Yano Cross-Reference

Yano uses the inclusive boundary:

```java
// Skip expired members. Committee terms are valid through the expiry epoch.
if (member.expiryEpoch() < currentEpoch) continue;
```

Source: `yano/ledger-state/src/main/java/com/bloxbean/cardano/yano/ledgerstate/governance/ratification/VoteTallyCalculator.java:234-235`

## Impact

This can affect indexer-visible governance output, including:

- committee vote tallying, because a valid hot-key vote can be ignored at the expiry boundary;
- committee acceptance ratio, because the active committee set can be computed incorrectly;
- committee minimum-size gating, because the committee can appear below minimum size one epoch too early;
- proposal status outcome, because ratification can fail/expire in Yaci Store while the ledger still counts the member as active.

## Suggested fix

Confirm that Yaci Store's `COMMITTEE_MEMBER.EXPIRED_EPOCH` stores the ledger `validUntil` epoch. If so, change the active committee member filter from:

```java
COMMITTEE_MEMBER.EXPIRED_EPOCH.gt(epoch)
```

to:

```java
COMMITTEE_MEMBER.EXPIRED_EPOCH.ge(epoch)
```

Keep the existing registration/deregistration epoch semantics unchanged:

- normal active-member lookup may include current-epoch registration/deregistration records;
- ratification lookup should still only include committee hot-key registrations that existed before entering the ratification epoch.

## Notes

The same active-member helper is used by both ratification membership resolution and committee vote collection. Changing the expiry predicate to inclusive can affect both paths, but that appears consistent with Cardano Ledger semantics: a committee member should remain active through the expiry epoch.

huylevt:
Closes #1022 
# Summary

  Fix committee member expiry boundary to match Cardano Ledger semantics. Committee membership is valid through the expiry epoch itself, so a member with `expiredEpoch == epoch` must still be treated as active.

  ## Problem

  Yaci Store previously filtered active committee members with:

  ```java
  COMMITTEE_MEMBER.EXPIRED_EPOCH.gt(epoch)
```
  This made committee expiry exclusive and excluded members one epoch too early.

  This also affected getActiveCommitteeMembersDetailsByEpoch(epoch), because that method uses the same active committee member predicate. As a result, committee members whose expiredEpoch matched the queried snapshot
  epoch were not returned for vote collection, even though Ledger treats them as still active.

  ## Changes

  - Changed committee member expiry filtering from EXPIRED_EPOCH > epoch to EXPIRED_EPOCH >= epoch.
  - Added storage tests for:
      - expiredEpoch == epoch remains active.
      - expiredEpoch < epoch is inactive.
      - Ratification lookup still excludes current-epoch hot-key registration.

  ## Not Affected

  The existing difference between the two lookup paths is preserved:

  - getActiveCommitteeMembersDetailsByEpoch(epoch) still resolves active members for vote collection at the snapshot epoch.
  - getActiveCommitteeMembersDetailsForRatificationByEpoch(epoch) still resolves active members for ratification at the current epoch.

  Only the expiry boundary changed; registration/deregistration timing behavior was not changed.

  ## Verification

  ./gradlew :stores:governance:test


satran004:
Reviewed — the boundary fix is correct per `cardano-ledger`: in `committeeAcceptedRatio` (`Ratify.hs`) a member is expired only when `currentEpoch > expiry`, so active ⇔ `EXPIRED_EPOCH >= epoch`. The stored `expiredEpoch` is the term `EpochNo` taken directly from `UpdateCommittee`, so `.ge(epoch)` matches the ledger exactly. Ran `:stores:governance:test` on this branch — all green.

One gap: the same off-by-one still exists in `GovernanceStatsService.getCommitteeStats()` (aggregates/governance-aggr), which counts a member as expired via `expiredEpoch <= epoch`. That's now inconsistent with this fix and with the ledger — should be `expiredEpoch < epoch`. Worth fixing in this PR (it's a two-character change) or filing a follow-up so it isn't lost.

Minor: the commented-out old implementation above `getActiveCommitteeMembersDetails` still shows the old `gt` predicate — good candidate for deletion here.

huylevt:
Thanks for the review. Addressed both points in 3f82c0175:                  
                                                                                                                                                                                                                                                                                                                                      
 - Updated GovernanceStatsService.getCommitteeStats() so a committee member is expired only when expiredEpoch < epoch, consistent with the ledger semantics and storage lookup.                                         
 - Removed the commented-out old implementation from CommitteeMemberStorageImpl.

satran004 (approved):
LGTM

PR #1025, opened 2026-07-21
Fix: require SPO approval for bootstrap SECURITY param changes

### LINKED ISSUE #1023: [Governance] Bootstrap phase: SECURITY parameter changes should still require SPO approval
## Summary

Yano and Yaci Store appear to treat bootstrap-phase parameter changes as committee-only once the committee threshold is met. Cardano Ledger resets DRep thresholds during the Conway bootstrap phase, but it does not remove the SPO threshold for parameter changes that touch the SECURITY parameter group.

This means a bootstrap/PV9 `ParameterChange` that modifies a SECURITY parameter should not be ratified by committee support alone. It still needs SPO acceptance.

## Cardano Ledger Rule

DRep thresholds are reset to zero during Conway bootstrap:

```haskell
thresholds@DRepVotingThresholds
  { ... }
    | hardforkConwayBootstrapPhase (pp ^. ppProtocolVersionL) = def
    | otherwise = pp ^. ppDRepVotingThresholdsL
```

Source: `cardano-ledger/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs:510-520`

For `ParameterChange`, the DRep threshold still exists, but is computed from the bootstrap-zero threshold set:

```haskell
ParameterChange _ ppu _ -> VotingThreshold $ pparamsUpdateThreshold thresholds ppu
```

Source: `cardano-ledger/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs:528-530`

SPO threshold selection is separate and does not disappear during bootstrap. If a parameter update affects any `SecurityGroup`, SPO voting is required:

```haskell
paramChangeThreshold ppu
  | any isSecurityRelevant (modifiedPPGroups ppu) =
      VotingThreshold pvtPPSecurityGroup
  | otherwise = NoVotingAllowed
```

Source: `cardano-ledger/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Internal.hs:391-394`

So the Ledger behavior for bootstrap SECURITY parameter changes is:

- Committee vote required.
- DRep vote auto-passes because the DRep threshold is zero.
- SPO vote required when the update touches SECURITY.

## Yano Behavior

Yano returns success immediately after the committee check during bootstrap:

```java
if (!checkCommittee(votes, committeeMembers, committeeThreshold, committeeState,
        currentEpoch, isBootstrapPhase, committeeMinSize)) return false;
if (isBootstrapPhase) return true;
```

Source: `yano/ledger-state/src/main/java/com/bloxbean/cardano/yano/ledgerstate/governance/ratification/RatificationEngine.java:391-393`

The SPO check for SECURITY groups is below that early return:

```java
if (ProtocolParamGroupClassifier.isSpoVotingRequired(affectedGroups)) {
    if (!checkSPO(votes, poolStakeDist, poolDRepDelegation,
            GovActionType.PARAMETER_CHANGE_ACTION, isBootstrapPhase, spoThresholds)) return false;
}
```

Source: `yano/ledger-state/src/main/java/com/bloxbean/cardano/yano/ledgerstate/governance/ratification/RatificationEngine.java:409-412`

Therefore, Yano accepts bootstrap SECURITY parameter changes with committee approval only, even when the Ledger requires SPO acceptance.

## Yaci Store Behavior

Yaci Store has the same semantic issue:

```java
if (context.isBootstrapPhase()) {
    isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD);
} else {
    List<ProtocolParamGroup> ppGroupChangeList = ProtocolParamUtil.getGroupsWithNonNullField(parameterChangeAction.getProtocolParamUpdate());
    VotingStatus dRepVotingResult = new DRepVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);

    if (ppGroupChangeList.contains(ProtocolParamGroup.SECURITY)) {
        VotingStatus spoVotingResult = new SPOVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
```

Source: `yaci-store/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java:43-50`

During bootstrap, the code never evaluates whether the parameter change touches SECURITY, and never checks SPO voting.

## Impact

This can affect indexer-visible governance output:

- proposal status can be marked ratified too early during bootstrap;
- Yano and Yaci Store can disagree with cardano-node for SECURITY parameter changes in protocol version 9 / Conway bootstrap environments.

## Suggested fix

For `ParameterChange` during bootstrap:

1. Compute affected protocol parameter groups before the bootstrap acceptance branch.
2. Keep DRep acceptance as automatic via zero bootstrap thresholds.
3. Keep committee acceptance required.
4. If the affected groups include SECURITY, require `SPOVotingEvaluator` / SPO threshold success.
5. If the affected groups do not include SECURITY, committee-only acceptance remains sufficient during bootstrap.

Apply the same semantic correction in both Yano and Yaci Store.

## Suggested verification

Use a Conway bootstrap/PV9 devnet or equivalent ledger fixture:

- Submit a `ParameterChange` touching a SECURITY parameter, such as `minFeeA`, with committee support but no SPO support. Expected outcome: not ratified; expires if no later SPO support is added.
- Submit the same action with committee support and enough SPO support. Expected outcome: ratified, assuming no other ratification gate blocks it.
- Submit a bootstrap-phase parameter change that does not touch SECURITY. Expected outcome: committee support is sufficient, because SPO voting is not required and DRep voting auto-passes in bootstrap.

huylevt:
## Summary
  - During Conway bootstrap phase, DRep voting thresholds are zeroed, but the SPO SECURITY-group threshold (pvtPPSecurityGroup) is unaffected and still applies.
  - ParameterChangeRatificationEvaluator previously accepted a bootstrap ParameterChange on committee support alone, skipping the SPO check entirely — even when the change touched a SECURITY parameter (e.g. minFeeA).
  - Fix: compute affected PParam groups before branching on bootstrap phase, and require SPO PASS_THRESHOLD whenever SECURITY is affected, regardless of bootstrap phase. Non-SECURITY bootstrap changes keep
  committee-only acceptance (unchanged).

  ## Test plan
  - Added evaluate_returnsContinue_whenBootstrap_securityParam_committeePasses_butSpoDoesNotSupport — reproduces the bug
  - Added evaluate_returnsAccept_whenBootstrap_securityParam_committeeAndSpoPass
  - Added evaluate_returnsAccept_whenBootstrap_nonSecurityParam_committeePassesAlone — no regression
  - ./gradlew :aggregates:governance-rules:test passes
  
 Closes #1023

### DIFF
--- a/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
+++ b/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
@@ -40,13 +40,21 @@ public class ParameterChangeRatificationEvaluator implements RatificationEvaluat
         final boolean isPreviousActionAsExpected = GovernanceActionUtil.isPrevActionAsExpected(parameterChangeAction.getType(), parameterChangeAction.getGovActionId(), lastEnactedGovActionId);
         boolean isAccepted;
 
+        List<ProtocolParamGroup> ppGroupChangeList = ProtocolParamUtil.getGroupsWithNonNullField(parameterChangeAction.getProtocolParamUpdate());
+        boolean spoVotingRequired = ppGroupChangeList.contains(ProtocolParamGroup.SECURITY);
+
         if (context.isBootstrapPhase()) {
             isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD);
+            // during bootstrap, DRep thresholds are zeroed, but the SPO SECURITY-group threshold
+            // is unaffected, so SPO approval is still required for SECURITY-group changes.
+            if (spoVotingRequired) {
+                VotingStatus spoVotingResult = new SPOVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
+                isAccepted = isAccepted && spoVotingResult.equals(VotingStatus.PASS_THRESHOLD);
+            }
         } else {
-            List<ProtocolParamGroup> ppGroupChangeList = ProtocolParamUtil.getGroupsWithNonNullField(parameterChangeAction.getProtocolParamUpdate());
             VotingStatus dRepVotingResult = new DRepVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
 
-            if (ppGroupChangeList.contains(ProtocolParamGroup.SECURITY)) {
+            if (spoVotingRequired) {
                 VotingStatus spoVotingResult = new SPOVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
                 if (ppGroupChangeList.size() == 1) {
                     isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD) && spoVotingResult.equals(VotingStatus.PASS_THRESHOLD);
--- a/aggregates/governance-rules/src/test/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluatorTest.java
+++ b/aggregates/governance-rules/src/test/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluatorTest.java
@@ -107,6 +107,135 @@ class ParameterChangeRatificationEvaluatorTest {
     }
 
     @Test
+    void evaluate_returnsContinue_whenBootstrap_securityParam_committeePasses_butSpoDoesNotSupport() {
+        VotingData votingData = VotingData.builder()
+                .committeeVotes(VotingData.CommitteeVotes.builder()
+                        .votes(Map.of(
+                                "hot1", com.bloxbean.cardano.yaci.core.model.governance.Vote.YES,
+                                "hot2", com.bloxbean.cardano.yaci.core.model.governance.Vote.YES))
+                        .build())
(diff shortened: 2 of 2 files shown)


satran004:
Looks good

A few minor observations (all pre-existing, non-blocking — fine as follow-ups):

1. `validateRequiredData` condition looks inverted: it requires `drepVotes != null` *in* bootstrap (where the DRep evaluator is never invoked) but validates nothing in the post-bootstrap branch, where missing DRep/SPO data fails closed via `INSUFFICIENT_DATA` and the action silently `CONTINUE`s instead of erroring.
2. The `ppGroupChangeList.size() == 1` special case in the post-bootstrap branch (skip DRep when only SECURITY is touched) is unreachable — every SECURITY field in `ProtocolParamUtil` also belongs to a DRep group, matching the ledger where DReps always vote on `ParameterChange`. Could be simplified away.
3. Trivial: `getGroupsWithNonNullField` is computed in both `evaluate()` and `validateRequiredData()`, and the new tests use fully-qualified `Vote.YES` instead of an import.

Sotatek-HuyLe3a:
Thanks, agreed. The missing-data validation is phase-inverted: DRep data should be required post-bootstrap, while SPO data should be required for SECURITY changes in either phase. The SECURITY-only branch is also unreachable with the current parameter grouping. I’ll track those two as follow-ups; missing data currently fails closed, so they are non-blocking. I’ll also clean up the duplicate group lookup and Vote import when touching this area.

satran004 (approved):

PR #1032, opened 2026-07-23
fix: make parameter change validation phase-aware

### LINKED ISSUE #1028: [Governance] Make ParameterChange validation phase-aware and simplify group handling
## Context

  PR #1025 fixed bootstrap-phase `ParameterChange` ratification so that changes affecting the `SECURITY` parameter group still require SPO approval.

  During review, several pre-existing issues were identified in `ParameterChangeRatificationEvaluator`:

  - Voting-data validation does not match the phase in which each evaluator runs.
  - The post-bootstrap `SECURITY`-only special case may be unreachable with correct protocol-parameter grouping.
  - Affected protocol-parameter groups are calculated more than once.

  These issues were non-blocking for #1025 because missing voting data currently fails closed, but they should be addressed separately.

  ## Problems

  ### 1. Voting-data validation is phase-inverted

  `validateRequiredData` currently requires DRep voting data during bootstrap, even though the DRep evaluator is not invoked in that phase.

  Conversely, post-bootstrap evaluation uses DRep voting data without validating it first. Missing data then reaches the evaluator, produces `INSUFFICIENT_DATA`, and causes the action to `CONTINUE` instead of failing
  through the intended validation path.

  The validation rules should be:

  - DRep voting data is required post-bootstrap.
  - SPO voting data is required whenever the parameter change affects `SECURITY`, in both bootstrap and post-bootstrap phases.
  - Existing committee-data requirements remain unchanged.

  ### 2. Review the `SECURITY`-only special case

  The post-bootstrap branch contains a special case based on:

  ```java
  ppGroupChangeList.size() == 1
```
  which skips DRep evaluation when only SECURITY is present.

  With correct Cardano Ledger parameter grouping, every security-relevant parameter should also belong to a DRep voting group. Therefore, this branch may be unreachable and could be removed.

  Before removing it, verify that all classifications in ProtocolParamUtil are correct, particularly the case tracked by #825 for minFeeRefScriptCostPerByte.

  ## Acceptance criteria

  - [ ] Bootstrap non-SECURITY changes do not require DRep or SPO voting data.
  - [ ] Bootstrap SECURITY changes require SPO voting data but not DRep voting data.
  - [ ] Post-bootstrap parameter changes require DRep voting data.
  - [ ] Post-bootstrap SECURITY changes require both DRep and SPO voting data.
  - [ ] Missing required data is handled by the intended validation path rather than silently falling through to INSUFFICIENT_DATA.
  - [ ] The SECURITY-only branch is either removed or retained with a documented, reachable use case.
  - [ ] Protocol-parameter classifications are verified against #825 before simplifying the branch.
  - [ ] Regression tests cover missing voting data across bootstrap/post-bootstrap and SECURITY/non-SECURITY combinations.
  - [ ] Existing tests from #1025 continue to pass.

  ## Verification

  ./gradlew :aggregates:governance-rules:test

  ## References

  - PR #1025
  - Review discussion: https://github.com/bloxbean/yaci-store/pull/1025#issuecomment-5044243323
  - Follow-up acknowledgement: https://github.com/bloxbean/yaci-store/pull/1025#issuecomment-5044650184
  - Related parameter-group classification issue: #825

### PR DESCRIPTION
**Pull request by huylevt**
## Summary

- Make `ParameterChange` voting-data validation phase-aware.
- Require DRep voting data only after the bootstrap phase.
- Require SPO voting data for SECURITY parameter changes in both phases.
- Remove the unreachable post-bootstrap SECURITY-only branch.
- Add regression coverage for missing voting data across bootstrap/post-bootstrap and SECURITY/non-SECURITY changes.

## Required voting data

| Phase | Non-SECURITY change | SECURITY change |
|---|---|---|
| Bootstrap | Committee | Committee + SPO |
| Post-bootstrap | Committee + DRep | Committee + DRep + SPO |

## Details

Previously, validation required DRep data during bootstrap even though DRep voting is not evaluated in that phase. Conversely, post-bootstrap evaluation used DRep data without validating its presence first, causing missing data to fall through as `INSUFFICIENT_DATA`.

The SECURITY-only branch was also removed. Every SECURITY parameter currently belongs to a DRep voting group as defined by the Cardano Ledger.

## Behavioral impact

Missing required voting data now fails through `validateRequiredData` with an `IllegalArgumentException` instead of silently producing `INSUFFICIENT_DATA` and returning `CONTINUE`.

The normal governance aggregation path already constructs the required voting-data components, so this primarily affects incomplete or malformed inputs.

## Verification

- `./gradlew :aggregates:governance-rules:test --tests com.bloxbean.cardano.yaci.store.governancerules.ratification.impl.ParameterChangeRatificationEvaluatorTest`
- `./gradlew :aggregates:governance-aggr:test`

Both completed successfully.

Fixes #1028

Related: #1025

### DIFF
--- a/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
+++ b/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
@@ -53,15 +53,13 @@ public class ParameterChangeRatificationEvaluator implements RatificationEvaluat
             }
         } else {
             VotingStatus dRepVotingResult = new DRepVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
+            isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD)
+                    && dRepVotingResult.equals(VotingStatus.PASS_THRESHOLD);
 
             if (spoVotingRequired) {
                 VotingStatus spoVotingResult = new SPOVotingEvaluator().evaluate(context.getVotingData(), votingEvaluationContext);
-                if (ppGroupChangeList.size() == 1) {
-                    isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD) && spoVotingResult.equals(VotingStatus.PASS_THRESHOLD);
-                } else
-                    isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD) && spoVotingResult.equals(VotingStatus.PASS_THRESHOLD) && dRepVotingResult.equals(VotingStatus.PASS_THRESHOLD);
-            } else
-                isAccepted = committeeVotingResult.equals(VotingStatus.PASS_THRESHOLD) && dRepVotingResult.equals(VotingStatus.PASS_THRESHOLD);
+                isAccepted = isAccepted && spoVotingResult.equals(VotingStatus.PASS_THRESHOLD);
+            }
         }
 
         if (context.isLastRatificationOpportunity()) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java [hunk 1]
```diff
@@ -79,16 +77,15 @@ public void validateRequiredData(RatificationContext context) {
...
 
-        if (context.isBootstrapPhase()) {
-            if (context.getVotingData().getDrepVotes() == null) {
-                throw new IllegalArgumentException("DRep votes are required for Parameter Change actions");
-            }
-            ParameterChangeAction parameterChangeAction = (ParameterChangeAction) context.getGovAction();
-            List<ProtocolParamGroup> ppGroupChangeList = ProtocolParamUtil.getGroupsWithNonNullField(parameterChangeAction.getProtocolParamUpdate());
+        if (!context.isBootstrapPhase() && context.getVotingData().getDrepVotes() == null) {
+            throw new IllegalArgumentException("DRep votes are required for Parameter Change actions");
+        }
```
satran004, line 82: @huylevt Will it throw exception in a devnet when no drep votes?

aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
```diff
@@ -79,16 +77,15 @@ public void validateRequiredData(RatificationContext context) {
...
 
-            if (ppGroupChangeList.contains(ProtocolParamGroup.SECURITY) && context.getVotingData().getSpoVotes() == null) {
-                throw new IllegalArgumentException("SPO votes are required for Parameter Change actions changing SECURITY parameters");
-            }
+        ParameterChangeAction parameterChangeAction = (ParameterChangeAction) context.getGovAction();
+        List<ProtocolParamGroup> ppGroupChangeList = ProtocolParamUtil.getGroupsWithNonNullField(parameterChangeAction.getProtocolParamUpdate());
+
+        if (ppGroupChangeList.contains(ProtocolParamGroup.SECURITY) && context.getVotingData().getSpoVotes() == null) {
+            throw new IllegalArgumentException("SPO votes are required for Parameter Change actions changing SECURITY parameters");
```
satran004, line 88: @huylevt same as previous question

aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/ratification/impl/ParameterChangeRatificationEvaluator.java
huylevt: No, not in the normal governance-aggr flow. “No DRep votes” is represented by a non-null DRepVotes snapshot, not by null.                                                                                                       
                                                                       
During bootstrap, governance-aggr creates a zero-valued DRep snapshot. Post-bootstrap, an empty vote list is still aggregated into a DRepVotes object with zero vote tallies and the remaining stake accounted as  doNotVoteStake.     
                                                                                                                                                                                                                                                                                                                                                                                                                             
Therefore, this validation only throws when a caller omits the required DRep data component entirely after bootstrap, which means the input is incomplete or malformed. A valid zero-vote snapshot is evaluated normally

satran004 (approved):

PR #1034, opened 2026-07-24
Fix SPO default votes per proposal

### LINKED ISSUE #1030: [Governance] Compute non-voting SPO default delegation per proposal
## Summary

`SPOVotingDataCollector.buildEpochAggregates` determines which stake pools are non-voters from SPO votes across **all proposals in the current evaluation batch**. It then computes `AlwaysAbstain` and `AlwaysNoConfidence` stake once and reuses those totals for every proposal.

This differs from `cardano-ledger`. The ledger stores SPO votes inside each `GovActionState` and applies a pool's default vote when that pool has no explicit vote for the **specific governance action currently being evaluated**.

Consequently, a pool that votes on proposal A but not proposal B can be excluded from Yaci Store's default-vote lookup for B. During post-bootstrap evaluation, this can produce:

- `AlwaysAbstain`: `No` instead of `Abstain`;
- `AlwaysNoConfidence` on a `NO_CONFIDENCE` action: `No` instead of `Yes`.

This behavior predates #1020 and was not introduced by that PR.

### SPO votes belong to one governance action

Each `GovActionState` owns its own `gasStakePoolVotes` map:

```haskell
gasId :: GovActionId
gasStakePoolVotes :: Map (KeyHash StakePool) Vote
```

Evidence:

- [`GovActionState` contains the action ID and its SPO vote map](https://github.com/IntersectMBO/cardano-ledger/blob/a624de4c8db7286a6c065da149679ea55f7d5629/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs#L219-L227)
- [`proposalsAddVote` updates the `GovActionState` selected by `GovActionId`](https://github.com/IntersectMBO/cardano-ledger/blob/a624de4c8db7286a6c065da149679ea55f7d5629/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Proposals.hs#L408-L429)

If pool P votes on A but not B, the ledger state is conceptually:

```text
A.gasStakePoolVotes = { P -> VoteYes }
B.gasStakePoolVotes = { }
```

### Default voting is evaluated against the current action

`spoAcceptedRatio` receives one `GovActionState` and looks up the pool in that action's `gasStakePoolVotes`:

```haskell
vote = Map.lookup poolId gasStakePoolVotes
```

Only when that lookup returns `Nothing` does the ledger apply HFI, bootstrap, or post-bootstrap default-vote behavior.

Evidence:

- [`spoAcceptedRatio` explicit-vote lookup and guard order](https://github.com/IntersectMBO/cardano-ledger/blob/a624de4c8db7286a6c065da149679ea55f7d5629/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs#L195-L226)
- [`defaultStakePoolVote` resolves `AlwaysAbstain` and `AlwaysNoConfidence`](https://github.com/IntersectMBO/cardano-ledger/blob/a624de4c8db7286a6c065da149679ea55f7d5629/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs#L560-L577)
- [Ledger property test for SPO accepted-ratio behavior](https://github.com/IntersectMBO/cardano-ledger/blob/a624de4c8db7286a6c065da149679ea55f7d5629/eras/conway/impl/testlib/Test/Cardano/Ledger/Conway/SPORatifySpec.hs#L56-L94)

```mermaid
flowchart TD
    Start["Tally one governance action"]
    Lookup{"Pool has an explicit vote<br/>for this action?"}
    Explicit["Use explicit Yes / No / Abstain"]
    HFI{"HardForkInitiation?"}
    HFINo["Count as No"]
    Bootstrap{"Bootstrap phase?"}
    BootAbstain["Count as Abstain"]
    Default["Resolve defaultStakePoolVote"]
    AA["AlwaysAbstain -> Abstain"]
    NC{"AlwaysNoConfidence<br/>and NoConfidence action?"}
    NCYes["Count as Yes"]
    No["Count as No"]

    Start --> Lookup
    Lookup -- Yes --> Explicit
    Lookup -- No --> HFI
    HFI -- Yes --> HFINo
    HFI -- No --> Bootstrap
    Bootstrap -- Yes --> BootAbstain
    Bootstrap -- No --> Default
    Default --> AA
    Default --> NC
    Default --> No
    NC -- Yes --> NCYes
    NC -- No --> No
```

## Current Yaci Store behavior

`VotingDataCollector.collectVotingDataBatch` retrieves SPO votes for all proposal IDs:

```java
var spoVotes = votingAggrService.getVotesBySPO(epoch, govActionIds);
```

It groups explicit votes by proposal, but it also passes the combined vote list to one shared epoch aggregate:

```text
spoVotes
├── groupVotesByGovAction(spoVotes)
└── buildEpochAggregates(spoVotes, epoch)
```

The resulting `SPOEpochAggregates` is reused for every proposal.

Evidence:

- [`collectVotingDataBatch` loads all votes, groups them, and creates one shared SPO aggregate](https://github.com/bloxbean/yaci-store/blob/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingDataCollector.java#L43-L104)
- [`getVotesBySPO` queries all supplied governance action IDs](https://github.com/bloxbean/yaci-store/blob/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java#L39-L99)

`buildEpochAggregates` constructs `poolsThatVoted` from that combined list:

```java
poolsThatVoted = spoVotes.stream()
        .map(VotingProcedure::getVoterHash)
        .collect(toSet());
```

Only pools absent from this batch-global set are considered for the `AlwaysAbstain` and `AlwaysNoConfidence` buckets.

Evidence:

- [`buildEpochAggregates` creates a batch-global `poolsThatVoted`](https://github.com/bloxbean/yaci-store/blob/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/SPOVotingDataCollector.java#L37-L92)

```mermaid
flowchart TD
    Proposals["Proposals A, B, C..."]
    Query["getVotesBySPO(epoch, all proposal IDs)"]
    AllVotes["Combined SPO votes"]

    Group["Group explicit votes by GovActionId"]
    AVotes["Explicit votes for A"]
    BVotes["Explicit votes for B"]

    Shared["buildEpochAggregates(all votes)"]
    GlobalSet["One batch-global poolsThatVoted"]
    Defaults["One shared default-delegation aggregate"]

    AData["collectSPOVotes(A votes, shared defaults)"]
    BData["collectSPOVotes(B votes, shared defaults)"]

    Proposals --> Query --> AllVotes
    AllVotes --> Group
    Group --> AVotes --> AData
    Group --> BVotes --> BData
    AllVotes --> Shared --> GlobalSet --> Defaults
    Defaults --> AData
    Defaults --> BData
```

The scope mismatch is:

```text
cardano-ledger: Did this pool vote on the action currently being tallied?
yaci-store:     Did this pool vote on any action in the evaluation batch?
```

## Conditions required to trigger the bug

The bug is observable only when **all** of the following conditions hold:

1. At least two proposals are evaluated in the same batch.
2. The same active stake pool votes on at least one proposal in that batch.
3. That pool does not vote on the target proposal being tallied.
4. The pool has a default delegation whose result differs from ordinary `No` for the target proposal:
   - `AlwaysAbstain` on a post-bootstrap non-HFI proposal; or
   - `AlwaysNoConfidence` on a post-bootstrap `NO_CONFIDENCE` proposal.
5. Evaluation is post-bootstrap.

### Behavior matrix

| Phase/action | Pool default | Ledger | Current Yaci classification | Diverges? |
|---|---|---|---|---|
| Bootstrap non-HFI | `AlwaysAbstain` | Abstain | `doNotVote` → Abstain | No |
| Bootstrap non-HFI | `AlwaysNoConfidence` | Abstain | `doNotVote` → Abstain | No |
| HFI in either phase | Any | No | `doNotVote` → No | No |
| Post-bootstrap non-HFI | `AlwaysAbstain` | Abstain | `doNotVote` → No | **Yes** |
| Post-bootstrap `NO_CONFIDENCE` | `AlwaysNoConfidence` | Yes | `doNotVote` → No | **Yes** |
| Post-bootstrap other action | `AlwaysNoConfidence` | No | `doNotVote` → No | No |
| Explicit vote on target proposal | Any | Explicit vote | Explicit vote | No |
| Pool votes on no proposal in batch | Special default | Default vote | Included in shared default bucket | No |

## Reproduction with concrete values

Post-bootstrap, evaluate proposals A and B in the same batch:

| Pool | Active stake | Default delegation | Vote on A | Vote on B |
|---|---:|---|---|---|
| P | 100 | `AlwaysAbstain` | Yes | No explicit vote |
| Q | 100 | Default `No` | Irrelevant | Yes |

Proposal B is a non-HFI action requiring an SPO threshold, for example a security `PARAMETER_CHANGE_ACTION`.

### Expected ledger result for B

For B, `gasStakePoolVotes` contains only Q. P is absent from B's vote map, so the ledger applies P's `AlwaysAbstain` default:

```mermaid
flowchart LR
    P["P: 100 stake"]
    PMissing["No explicit vote on B"]
    PDefault["DefaultAbstain"]
    Abstain["B Abstain += 100"]

    Q["Q: 100 stake"]
    QYes["Explicit Yes on B"]
    Yes["B Yes += 100"]

    P --> PMissing --> PDefault --> Abstain
    Q --> QYes --> Yes
```

| Ledger category for B | Stake |
|---|---:|
| Yes | 100 |
| Abstain | 100 |
| No | 0 |

```text
accepted ratio = Yes / (Total - Abstain)
               = 100 / (200 - 100)
               = 1.0
```

### Actual Yaci Store result for B

The combined batch contains P's vote on A, so:

```text
poolsThatVoted = { P, Q }
```

P is excluded from the batch-global non-voter set. Its 100 stake is therefore missing from the shared `AlwaysAbstain` bucket. When B is aggregated, that stake becomes the remaining `doNotVoteStake`.

`computeSPOTallies` does not add post-bootstrap `doNotVoteStake` to Yes or Abstain. `totalNoStake` is calculated as the remaining total, so P is counted as No.

Evidence:

- [`collectSPOVotes` calculates `doNotVoteStake` as the remaining stake](https://github.com/bloxbean/yaci-store/blob/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/SPOVotingDataCollector.java#L95-L122)
- [`computeSPOTallies` makes remaining post-bootstrap stake implicit No](https://github.com/bloxbean/yaci-store/blob/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982/aggregates/governance-rules/src/main/java/com/bloxbean/cardano/yaci/store/governancerules/voting/VoteTallyCalculator.java#L61-L101)

```mermaid
flowchart LR
    PVoteA["P voted on A"]
    Global["P is in batch-global poolsThatVoted"]
    Excluded["P excluded from non-voter default lookup"]
    Missing["P missing from shared AlwaysAbstain stake"]
    BNoVote["P has no explicit vote on B"]
    DNV["P stake becomes B.doNotVote"]
    No["Post-bootstrap remainder becomes No"]

    PVoteA --> Global --> Excluded --> Missing --> BNoVote --> DNV --> No
```

| Category for B | Ledger | Current Yaci |
|---|---:|---:|
| Yes | 100 | 100 |
| Abstain | 100 | 0 |
| No | 0 | 100 |
| Accepted ratio | 1.0 | 0.5 |

With a threshold of `0.6`, the ledger passes B while Yaci Store fails it.

## Impact

The incorrect classification can affect:

- SPO Yes/No/Abstain totals;
- SPO accepted/approval ratio;
- proposal ratification result when the ratio crosses a threshold;
- persisted `gov_action_proposal_status.voting_stats`;
- persisted proposal `status` when the voting result changes.

Historical post-bootstrap rows may require assessment or reprocessing after the fix.

## Expected behavior

For every proposal:

1. Determine the explicit SPO voter set for that proposal.
2. Apply HFI/bootstrap/default rules only to pools without an explicit vote on that proposal.
3. Let an explicit vote override the pool's default delegation for that proposal.
4. Do not let a vote on proposal A affect default classification on proposal B.

## Suggested approach

Keep storage access efficient by loading an epoch-level pool snapshot once:

```text
poolId -> {
    activeStake,
    defaultVote
}
```

Then classify and sum that snapshot independently for each proposal using the proposal's explicit voter set.

```mermaid
flowchart TD
    Snapshot["Epoch snapshot:<br/>pool -> stake + defaultVote"]
    Votes["All SPO votes"]
    Group["Group votes by GovActionId"]
    ASet["A explicit-voter set"]
    BSet["B explicit-voter set"]
    AggA["Aggregate A independently"]
    AggB["Aggregate B independently"]

    Snapshot --> AggA
    Snapshot --> AggB
    Votes --> Group
    Group --> ASet --> AggA
    Group --> BSet --> AggB
```

This preserves batch query efficiency without sharing proposal-dependent classification.

## Acceptance criteria

- Default-vote classification is performed independently for each proposal.
- A pool voting on A does not suppress its default vote on B.
- An explicit vote overrides default delegation on the same proposal.
- No pool stake appears in more than one category for a proposal.
- The categories reconcile with total active stake:

  ```text
  explicitYes
  + explicitNo
  + explicitAbstain
  + defaultAlwaysAbstain
  + defaultAlwaysNoConfidence
  + ordinaryDoNotVote
  = totalActiveStake
  ```

- Bootstrap non-HFI and HFI behavior remain unchanged.
- The implementation does not introduce one storage query per pool per proposal.

## Regression tests

1. **AlwaysAbstain is applied per proposal**
   - P is delegated to `AlwaysAbstain`.
   - P votes Yes on A but does not vote B.
   - B includes P in `delegateToAutoAbstainDRepStake`.
   - B does not include P in `doNotVoteStake`.
   - Post-bootstrap B counts P as Abstain.

2. **AlwaysNoConfidence is applied per proposal**
   - P is delegated to `AlwaysNoConfidence`.
   - P votes on A but not on `NO_CONFIDENCE` proposal B.
   - B includes P in `delegateToNoConfidenceDRepStake`.
   - Post-bootstrap B counts P as Yes.

3. **Explicit vote overrides the default**
   - P is delegated to `AlwaysAbstain`.
   - P explicitly votes Yes on B.
   - B counts P only as explicit Yes.

4. **Unrelated behavior remains unchanged**
   - Bootstrap non-HFI non-voters remain Abstain.
   - HFI non-voters remain No in both phases.
   - Post-bootstrap `AlwaysNoConfidence` remains No for actions other than `NO_CONFIDENCE`.

## Verification references

Verified against:

- `yaci-store` commit [`76e6f2964532eff73ed7309ccf9bb9b1f8ad0982`](https://github.com/bloxbean/yaci-store/commit/76e6f2964532eff73ed7309ccf9bb9b1f8ad0982);
- `cardano-ledger` commit [`a624de4c8db7286a6c065da149679ea55f7d5629`](https://github.com/IntersectMBO/cardano-ledger/commit/a624de4c8db7286a6c065da149679ea55f7d5629).

huylevt:
## Summary

- Compute SPO default votes per governance proposal instead of sharing a batch-global non-voter classification across all proposals evaluated in the same epoch.
- Preserve pool-level `AlwaysAbstain` and `AlwaysNoConfidence` stake in the epoch snapshot so an explicit vote suppresses the default only for the proposal that received that vote.
- Keep the previous aggregate accessors and deprecated `buildEpochAggregates(List<VotingProcedure>, int)` overload for compatibility.
- Add regression coverage for both default delegations, explicit-vote precedence, multi-row pool stake aggregation, and post-bootstrap Hard Fork Initiation behavior.

## Root cause

`VotingDataCollector` loads SPO votes for all proposals being evaluated in an epoch and groups the explicit votes by proposal. Previously, `SPOVotingDataCollector.buildEpochAggregates` built one `poolsThatVoted` set from the combined vote list and excluded those pools from the default-delegation lookup for every proposal in the batch.

This meant that a pool which voted on proposal A but not proposal B was treated as having voted for both when the shared epoch aggregate was built. On proposal B, Yaci Store therefore put the pool in the generic `doNotVote` bucket instead of applying the pool reward account's default DRep delegation.

The ledger lookup is action-specific: `spoAcceptedRatio` first checks for an explicit vote in the current action's `gasStakePoolVotes`; only when that vote is absent does it apply the pool's default delegation.

## Fix

The epoch snapshot now resolves `AlwaysAbstain` and `AlwaysNoConfidence` stake for all active pools and retains that stake by pool ID. While collecting one proposal, the collector removes only pools that explicitly voted on that proposal before summing the applicable default stake.

As a result:

- A vote on proposal A no longer suppresses the pool's default on proposal B.
- An explicit vote on proposal B still takes precedence over the pool's default on B.
- `AlwaysAbstain` stake is classified as `Abstain` where required by the post-bootstrap ledger rules.
- `AlwaysNoConfidence` stake is classified as `Yes` for `NO_CONFIDENCE` actions.
- Hard Fork Initiation remains phase-independent and continues to treat non-explicit/default stake according to its dedicated rule.

## Historical network impact

The read-only database analysis found stale historical `voting_stats` in the following evaluation epochs:

| Network | Affected evaluation epochs | Required correction | Ratification result |
|---|---|---|---|
| Mainnet | `563`, `573–580`, `600–618`, `638–645` | `AlwaysAbstain`: `No → Abstain` | No threshold crossing found. The only threshold-relevant epochs were `574`, `601`, and `607`. |
| Preprod | `278–284`, `293` | `AlwaysAbstain`: `No → Abstain` | No status impact; all affected targets were treasury withdrawals. |
| Preview | `770–790`, `802`, `815–843`, `1007–1060` | `AlwaysAbstain`: `No → Abstain`; `AlwaysNoConfidence` on `NO_CONFIDENCE`: `No → Yes` | No fail-to-pass or pass-to-fail crossing found. |

The code change affects newly computed results only. Existing `gov_action_proposal_status.voting_stats` rows are not rewritten automatically.

### Automated verification

- `./gradlew :aggregates:governance-aggr:test --tests com.bloxbean.cardano.yaci.store.governanceaggr.service.SPOVotingDataCollectorTest`
- `./gradlew :aggregates:governance-aggr:build :aggregates:governance-rules:build`
- `./gradlew build`
- `git diff --check`

All commands completed successfully. The workspace build completed 683 tasks. The repository's `e2e-tests:test` task remained skipped by its existing configuration.

### Preprod re-sync verification

- [x] Deploy the fixed build to the Preprod verification environment.
- [x] Reprocess proposal-status evaluation epochs 278–293, or re-sync from a point before evaluation epoch 278 so those rows are rebuilt by the normal aggregation pipeline.
- [x] Confirm the 24 affected contributions move from SPO `No` to `Abstain`.
- [x] Confirm the five affected treasury withdrawal proposals retain their existing statuses.
- [x] Sync through the current Preprod tip and confirm governance aggregation completes without errors or material epoch-processing regression.

### Preview re-sync verification

- [x] Deploy the fixed build to the Preview verification environment.
- [x] Reprocess proposal-status evaluation epochs 770–1060, or re-sync from a point before evaluation epoch 770 so those rows are rebuilt by the normal aggregation pipeline.
- [x] Confirm affected `AlwaysAbstain` contributions move from SPO `No` to `Abstain`.
- [x] Confirm affected `AlwaysNoConfidence` contributions on `NO_CONFIDENCE` proposals move from SPO `No` to `Yes`.
- [x] Confirm the reconstructed pass/fail results remain unchanged for all SPO-threshold-relevant contexts.
- [x] Sync through the current Preview tip and confirm governance aggregation completes without errors or material epoch-processing regression.

## Data and rollout considerations

- Add a release-note entry explaining that previously persisted governance voting statistics may require reprocessing.
- Use the Preprod and Preview re-sync checks above before deciding whether to perform targeted Mainnet reprocessing.
- Mainnet reprocessing should be planned separately because of the production data volume; the current analysis supports correcting historical statistics but found no ratification status requiring repair.

Fixes #1030


satran004:
Looks good.

The per-proposal exclusion matches cardano-ledger's `spoAcceptedRatio` semantics, downstream tallies (`VoteTallyCalculator.computeSPOTallies`) consume the fields per proposal so HFI and bootstrap handling are unaffected, and the delegation lookup's latest-row semantics rule out double counting between the abstain and no-confidence buckets. The added `.distinct()` also fixes a latent cross-batch double-count in the old code.

A few minor suggestions:

1. **`defaultStakeForNonVoters` legacy fallback is subtle and undocumented.** When `stakeByPool` is empty it returns the aggregate value, which silently reverts to the old batch-global semantics for an `SPOEpochAggregates` built via the 4-arg compatibility constructor. In the normal path this branch only ever returns zero, so the `isEmpty()` check reads like an optimization when it's really a compatibility shim. Add a comment saying so — or, since the only internal caller (`VotingDataCollector`) is already migrated, consider dropping the 4-arg constructor and the deprecated overload entirely.

2. **The deprecated `buildEpochAggregates(List, int)` silently ignores its `spoVotes` argument.** Any external caller reading the aggregate accessors (e.g. `delegateToAutoAbstainDRepStake()`) directly now gets *all* default-delegated stake instead of non-voting stake. If the overload is kept, the `@Deprecated` javadoc should call out this behavior change explicitly.

3. **Test-isolation risk in the new IT.** `spoAlwaysAbstainDefault_shouldBeAppliedPerProposal` uses `accountAt(21)` as the pool owner — the same account the existing hard-fork-vs-update-committee edge test uses for its AlwaysAbstain pool. If both tests run against one DevKit instance, the leftover pool from the other test shares the reward account (and its AlwaysAbstain delegation), and residual snapshot stake on that old pool could break the exact-equality assertion `spoTotalAbstainStake == poolPStake`. A dedicated account index would be cheap insurance.

4. **PR description slightly overstates new coverage.** It mentions regression coverage for "post-bootstrap Hard Fork Initiation behavior," but the three new unit tests don't include an HFI case — that behavior lives in `VoteTallyCalculator` and the pre-existing IT. Worth either adjusting the description or adding a small unit test asserting the per-proposal fields feed HFI tallies unchanged.

huylevt:
@satran004   Thanks for the detailed review. I addressed suggestions 1, 2, and 4 in the follow-up commit `8bfa3e043`.

  1. Added a comment explaining that the empty `stakeByPool` branch is a compatibility fallback for aggregate-only snapshots created through the four-argument constructor. Snapshots produced by the normal
  `buildEpochAggregates(int)` path have a corresponding aggregate of zero when the map is empty.

  2. Expanded the deprecated overload's Javadoc to clarify that `spoVotes` is intentionally ignored and that its aggregate accessors contain stake from all matching default-delegated pools. Proposal-specific non-voter exclusion
  happens later in `collectSPOVotes`.

  3. I rechecked the test-isolation concern. Although both tests use `accountAt(21)`, `GovernanceRuleEdgeIT` uses `@DirtiesContext(AFTER_EACH_TEST_METHOD)`. Its initializer recreates the DevNet with overwrite enabled, which
  stops the existing cluster and deletes its previous data directory. Therefore, residual pool state cannot carry between these tests under the standard test lifecycle, so I kept the account index unchanged.

  4. Adjusted the PR description to state that the new tests cover proposal-scoped default delegations, explicit-vote precedence, and multi-row stake aggregation, while the existing bootstrap and post-bootstrap HFI tally tests
  continue to cover the downstream HFI behavior.

satran004 (approved):

PR #1046, opened 2026-07-29
Fix full table scan in findLatestBlock

### LINKED ISSUE #1041: Blockfrost /blocks/latest performs full table scan due to DESC NULLS LAST ordering
## Description

`GET /api/v1/blockfrost/blocks/latest` becomes increasingly slow as the `block` table grows, even though `idx_block_number` exists.

This is reproducible on `v3.0.0-beta3`, and the same query is present in the current source.

## Reproduction

```bash
curl -X GET \
  'https://yaci-playground.cf-app.org/api/v1/blockfrost/blocks/latest' \
  -H 'accept: */*'
```

On a mainnet database with about 13.7 million blocks:

- Public endpoint: approximately 8.5–8.9 seconds
- PostgreSQL execution: approximately 7.5–8.9 seconds
- `block` heap size: approximately 12 GB
- Buffers read by one request: approximately 1.59 million 8 KB pages

## Root cause

`BFBlocksStorageReaderImpl.findLatestBlock()` currently uses:

```java
return blockBaseSelect(block, nextBlockField, confirmationsField, DSL.trueCondition())
        .orderBy(block.NUMBER.desc().nullsLast())
        .limit(1)
        .fetchOptional(record -> toBlockRow(record, block, nextBlockField, confirmationsField));
```

The schema declares `block.number` as nullable and the existing index is:

```sql
CREATE INDEX idx_block_number ON block(number);
```

A backward scan of this index provides `DESC NULLS FIRST`, so it cannot satisfy the explicit `DESC NULLS LAST` ordering. PostgreSQL therefore performs a sequential scan of the entire table followed by a top-N sort.

Relevant plan excerpt:

```text
Limit
  -> Sort
       Sort Key: b.number DESC NULLS LAST
       Sort Method: top-N heapsort
       -> Seq Scan on block b
            actual rows=13,735,493
Buffers: shared hit=4,727 read=1,591,076
Execution Time: 7,478.900 ms
```

Statistics reported no actual null values for `block.number`, but that does not allow the planner to use the index for the explicit null ordering because the column remains nullable in the schema.

## Proposed fix

Exclude null block numbers and remove the explicit null ordering:

```java
return blockBaseSelect(
        block,
        nextBlockField,
        confirmationsField,
        block.NUMBER.isNotNull()
    )
    .orderBy(block.NUMBER.desc())
    .limit(1)
    .fetchOptional(record ->
        toBlockRow(record, block, nextBlockField, confirmationsField));
```

The equivalent SQL was tested against the same live database. It reused the existing `idx_block_number` index:

```text
Limit
  -> Index Scan Backward using idx_block_number on block b
       Index Cond: (number IS NOT NULL)
Buffers: shared hit=11
Execution Time: 0.182 ms
```

This reduces database execution from several seconds to sub-millisecond without adding another index.

## Database-only workaround

For deployments that cannot immediately upgrade/rebuild, a matching index avoids the full scan:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_block_number_desc_nulls_last
    ON block (number DESC NULLS LAST);
```

The application-level fix is preferable because it reuses the existing index.

Sotatek-HenryDo:
## Summary
- `findLatestBlock()` ordered by `block.number.desc().nullsLast()` with no `WHERE` filter. Since `number` is nullable, a backward index scan on `idx_block_number` naturally yields `NULLS FIRST`, which can't satisfy an explicit `NULLS LAST`, forcing a full sequential scan + sort on large tables.
- Fix: filter `block.NUMBER.isNotNull()` and order by a plain `.desc()` so Postgres can reuse the existing index via a backward index scan.

Fixes #1041

## Test plan
Verified with `EXPLAIN (ANALYZE, BUFFERS)` against a synced mainnet DB (13.7M blocks):
- Before: `Parallel Seq Scan` + Sort, ~5.9s, ~1.59M buffer pages read
- After: `Index Scan Backward` on `idx_block_number`, ~0.07ms, 4 buffer pages read

Also verified end-to-end on the live `/api/v1/blockfrost/blocks/latest` endpoint on the same mainnet node:
- Before: ~7.4-7.5s response time
- After: ~6ms response time (steady state)

### DIFF
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
@@ -41,8 +41,8 @@ public class BFBlocksStorageReaderImpl implements BFBlocksStorageReader {
         var nextBlockField = nextBlockField(block);
         var confirmationsField = confirmationsField(block);
 
-        return blockBaseSelect(block, nextBlockField, confirmationsField, DSL.trueCondition())
-                .orderBy(block.NUMBER.desc().nullsLast())
+        return blockBaseSelect(block, nextBlockField, confirmationsField, block.NUMBER.isNotNull())
+                .orderBy(block.NUMBER.desc())
                 .limit(1)
                 .fetchOptional(record -> toBlockRow(record, block, nextBlockField, confirmationsField));
     }


Sotatek-HuyLe3a (approved):

satran004 (approved):
Thanks @Sotatek-HenryDo

Issue #1049, opened 2026-07-31
Blockfrost /txs/{hash} performs an expensive skip scan on tx_input.spent_tx_hash


Sotatek-HuyLe3a:
## Description

`GET /api/v1/blockfrost/txs/{hash}` becomes slow on a mainnet-sized database because the transaction lookup calculates `input_count` using only `tx_input.spent_tx_hash`.

Current code in `BFTransactionStorageReaderImpl.findTransactionByHash()`:

```java
Field<Integer> inputCount = DSL.selectCount()
        .from(TX_INPUT)
        .where(TX_INPUT.SPENT_TX_HASH.eq(txHash))
        .asField("input_count");
```

The same predicate is also used by `findTxUtxos()`:

```java
.where(TX_INPUT.SPENT_TX_HASH.eq(txHash))
```

The database does not currently have an index whose leading column is `spent_tx_hash`. It has the following composite index instead:

```sql
CREATE INDEX idx_tx_input_spent_at_block_spent_tx_hash
    ON tx_input (spent_at_block, spent_tx_hash);
```

Because `spent_tx_hash` is the second column, PostgreSQL 18 uses an expensive index skip scan.

Related performance documentation:

- https://github.com/bloxbean/yaci-store/wiki/Blockfrost-API-Performance-Baselines
- https://github.com/bloxbean/yaci-store/wiki/Additional-BF-Indexes

## Reproduction

Representative transaction:

```text
0d1d90d5a5e7a694a7d126932afbd6cddd824ef389bf068f7f0dc376edff58f9
```

Query:

```sql
EXPLAIN (
    ANALYZE,
    BUFFERS,
    WAL,
    TIMING OFF,
    SUMMARY ON
)
SELECT count(*)
FROM tx_input
WHERE spent_tx_hash =
    '0d1d90d5a5e7a694a7d126932afbd6cddd824ef389bf068f7f0dc376edff58f9';
```

Environment:

- PostgreSQL 18.1
- `tx_input`: approximately 341.7 million rows / 160 GB
- `transaction`: approximately 122.8 million rows
- `address_utxo`: approximately 352.8 million rows

Observed plan:

```text
Aggregate
  -> Index Only Scan using idx_tx_input_spent_at_block_spent_tx_hash
       Index Cond: spent_tx_hash = ...
       actual rows=1
       Heap Fetches: 1
       Index Searches: 101332
       Buffers: shared hit=398862 read=1817710

Execution Time: 12650.280 ms
```

[...]

PR #1050, opened 2026-07-31
[Blockfrost API] Optimize Blockfrost transaction input lookup

Sotatek-HuyLe3a:
## Summary

Improves the performance of the Blockfrost-compatible transaction endpoints by constraining `tx_input` lookups with both the consuming block number and transaction hash.

This allows the queries to use the existing `tx_input (spent_at_block, spent_tx_hash)` index efficiently, without adding another index or changing the database schema.

Related to #1049 .

## Problem

`GET /api/v1/blockfrost/txs/{hash}` calculated the input count using only `spent_tx_hash`:

```sql
WHERE tx_input.spent_tx_hash = ?
```

On a mainnet database with more than 341 million `tx_input` rows, PostgreSQL had to perform an index skip scan over many `spent_at_block` values because `spent_tx_hash` is the second column of the existing composite index.

The same lookup pattern was also used when loading inputs for `GET /api/v1/blockfrost/txs/{hash}/utxos`.

## Changes

- Correlate the transaction input count with `transaction.block`:

```sql
WHERE tx_input.spent_at_block = transaction.block
  AND tx_input.spent_tx_hash = ?
```

- Use the transaction's block number when loading its consumed inputs for the UTXO endpoint:

```sql
WHERE tx_input.spent_at_block = ?
  AND tx_input.spent_tx_hash = ?
```

No Flyway migration or additional database index is required.

## Correctness

All inputs consumed by a transaction are recorded with the block in which that transaction was included. Therefore, adding `spent_at_block` narrows the search to the correct index range without changing the selected inputs.

This also preserves invalid-transaction behavior. For an invalid transaction, `tx_input` contains the collateral inputs that were consumed instead of the regular transaction inputs. Those collateral inputs are consumed in the invalid transaction's block, so the same block predicate applies.

For valid transactions, regular consumed inputs continue to come from `tx_input`; collateral and reference inputs continue to be handled by the existing separate branches.

### Database query plan

Verified with `EXPLAIN (ANALYZE, BUFFERS)` against a PostgreSQL mainnet database:

| Query | Before | After |
|---|---:|---:|
| Isolated `tx_input` count | ~12.65 s | ~0.126 ms |
| Full transaction aggregate after the rewrite | N/A | ~64.5 ms |

The rewritten count query used the existing composite index with a single targeted index search instead of skip-scanning across block values.

### Deployed API

Verified using Docker image `bloxbean/yaci-store:fix-1049` and a different mainnet transaction:

```text
ee578fba7424bee97b4ad49f52b0480c8fd0f238c26278cf0c94a4d446a26b4a
```

| Endpoint | First request | 10-request average | Min | Max | Result |
|---|---:|---:|---:|---:|---|
| `/api/v1/blockfrost/txs/{hash}` | 77.7 ms | 26.5 ms | 19.2 ms | 41.6 ms | HTTP 200 |
| `/api/v1/blockfrost/txs/{hash}/utxos` | 74.6 ms | 21.3 ms | 16.3 ms | 26.9 ms | HTTP 200 |

The transaction response matched the requested hash and returned block height `13748070`, `valid_contract: true`, and `utxo_count: 5`. The UTXO response returned the expected regular, collateral, and reference input classifications together with the transaction outputs.

### Build

```bash
./gradlew :extensions:blockfrost:blockfrost-transaction:compileJava
```

Result: `BUILD SUCCESSFUL` with Java 21.

## Database impact

- No new index
- No schema migration
- No data migration
- No changes to stored transaction data

### DIFF
--- a/extensions/blockfrost/transaction/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/transaction/storage/impl/BFTransactionStorageReaderImpl.java
+++ b/extensions/blockfrost/transaction/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/transaction/storage/impl/BFTransactionStorageReaderImpl.java
@@ -86,7 +86,9 @@ public class BFTransactionStorageReaderImpl implements BFTransactionStorageReade
                 .asField("asset_mint_count");
 
         Field<Integer> inputCount = DSL.selectCount()
-                .from(TX_INPUT).where(TX_INPUT.SPENT_TX_HASH.eq(txHash))
+                .from(TX_INPUT)
+                .where(TX_INPUT.SPENT_AT_BLOCK.eq(TRANSACTION.BLOCK))
+                .and(TX_INPUT.SPENT_TX_HASH.eq(txHash))
                 .asField("input_count");
 
         Field<Integer> outputCount = DSL.selectCount()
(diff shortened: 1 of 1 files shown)


satran004 (approved):
Thanks @huylevt . LGTM

Sotatek-HuyLe3a:
Possible follow-up: BFAccountStorageReaderImpl.java:835 filters TX_INPUT.SPENT_TX_HASH.in(txHashes) while joined to TRANSACTION — the same correlation trick (SPENT_AT_BLOCK =
TRANSACTION.BLOCK) could likely speed that account endpoint up too. Worth a separate issue, not this PR.

PR #1054, opened 2026-08-02
fix: ignore stale DRep votes after unregistration

### LINKED ISSUE #1048: DRep vote from before re-registration is incorrectly counted in voting tally
## Summary

`VotingAggrService.getVotesByDRep()` selects a DRep's vote for a governance action using `MAX(slot)` combined with a check that the DRep currently exists in `drep_dist` at the ratification snapshot epoch. It never checks whether that vote occurred **after** the DRep's most recent unregistration.

As a result, when a DRep votes on a proposal, unregisters, and later **re-registers the same credential** without casting a new vote on that proposal, the stale pre-unregistration vote is picked up again and counted toward the tally — even though Conway ledger semantics deleted that vote the moment the `UnRegDRepTxCert` was processed, and re-registration does not restore it.

This produces incorrect `drep_*_vote_stake` / `drep_do_not_vote_stake` / `drep_total_*_stake` / `drep_approval_ratio` values for any proposal affected by this lifecycle pattern.

huylevt:
## Summary

  Fix DRep vote tallying so votes cast before a DRep unregistration are not revived when the same DRep credential is registered again.

  Fixes #1048.

  ## Root cause

  Cardano ledger state clears a DRep's votes when the DRep unregisters. Re-registering the same credential does not restore those votes.

  Yaci Store previously selected the latest stored vote and only checked whether the DRep existed in the snapshot distribution. Once the DRep registered again, this eligibility check passed and the stale pre-
  unregistration vote was incorrectly included in governance voting statistics, causing mismatches with AdaStat and ledger state.

  ## Changes

  - Exclude DRep votes invalidated by an unregistration at or before the requested snapshot epoch.
  - Preserve votes cast after the DRep registers again.
  - Select the latest effective vote deterministically using slot, transaction index, and vote index.
  - Match DRep credential types when checking registration and distribution records.
  - Add unit coverage for:
    - stale votes after unregister/re-register;
    - new votes cast after re-registration;
    - unregistrations occurring after the requested snapshot epoch.
  - Extend the DevKit E2E lifecycle test to:
    - re-register and restore the DRep delegation;
    - verify that the old vote remains cleared;
    - verify that a new post-registration vote is counted.
  - Update the governance E2E test-case documentation.

  Raw historical vote rows remain unchanged; lifecycle filtering is applied when determining effective votes. No database migration is required.

  ## Testing

  - [x] `./gradlew :aggregates:governance-aggr:test --tests com.bloxbean.cardano.yaci.store.governanceaggr.service.VotingAggrServiceTest --console=plain`
  - [x] `./gradlew :e2e-tests:compileTestJava`
  - [x] `./gradlew :e2e-tests:test --tests com.bloxbean.cardano.yaci.store.test.e2e.gov.GovernanceDRepLifecycleVoteTallyIT --console=plain`
    - 1 test, 0 failures, 0 errors

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java
@@ -1,5 +1,7 @@
 package com.bloxbean.cardano.yaci.store.governanceaggr.service;
 
+import com.bloxbean.cardano.yaci.core.model.certs.CertificateType;
+import com.bloxbean.cardano.yaci.core.model.certs.StakeCredType;
 import com.bloxbean.cardano.yaci.core.model.governance.GovActionId;
 import com.bloxbean.cardano.yaci.core.model.governance.Vote;
 import com.bloxbean.cardano.yaci.core.model.governance.VoterType;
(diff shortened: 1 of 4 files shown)


satran004 (approved):

huylevt:
Worth a follow-up look (out of scope)

  getVotesBySPO and the committee path still use the old max(slot) pattern — they share the same latent same-slot duplicate issue this PR fixed for DReps, and it may be worth confirming
  whether analogous lifecycle events (pool retirement, committee resignation) need similar treatment for tally parity.

PR #962, opened 2026-05-26
feat(admin-cli): add generic index apply command

Sotatek-HuyLe3a:
## Summary

  This PR add  a generic optional index command that can apply one or more named index definitions and external index definition files. 

  The existing `apply-indexes` and `apply-extra-indexes` commands remain unchanged.

  ## Changes

  - Add `apply-optional-indexes` command for optional index definitions.
  - Enhance `IndexLoader` to support:
    - multiple index definition files
    - named index definitions resolved from classpath resources
    - external resources via `file:`, `classpath:`, `http:`, and `https:`
  - Keep shorthand named index usage ergonomic by resolving names without extension as `.yml`.
  - Add `IndexLoader` tests for multi-file loading, `.yml` shorthand resolution, external `file:` loading, and missing file handling.

  ## Command Usage

  ```bash
  ## Apply BF Compatibility Indexes
  apply-optional-indexes blockfrost-index

  ## Apply multiple named index definitions:

  apply-optional-indexes blockfrost-index,asset-ext-index

  ## Apply an external index definition file:

  apply-optional-indexes --index-files file:///opt/yaci-store/custom-index.yml

  ## Apply multiple external index definition files:

  apply-optional-indexes --index-files file:///opt/index-a.yml,file:///opt/index-b.yml

  ## Apply a named index definition together with an external file:

  apply-optional-indexes blockfrost-index --index-files file:///opt/yaci-store/custom-index.yml
```
  ## Notes

  - External resource paths should be passed through --index-files.
  - Existing commands still work:
      - apply-indexes
      - apply-extra-indexes

PR #1059, opened 2026-08-04
Fix full table scan in block addresses endpoint

### LINKED ISSUE #1044: Blockfrost block addresses endpoint scans address_utxo due to block/block_hash index mismatch
## Description

`GET /api/v1/blockfrost/blocks/{hash_or_number}/addresses` performs a sequential scan of the full `address_utxo` table and does not return within a reasonable time on a mainnet database.

This is reproducible on `v3.0.0-beta3`, and the same query remains in the current source.

## Reproduction

```bash
curl -X GET \
  'https://yaci-playground.cf-app.org/api/v1/blockfrost/blocks/13732068/addresses?count=100&page=1' \
  -H 'accept: */*'
```

Observed against a mainnet database:

- A block with 57 transactions timed out after 45 seconds without receiving response bytes.
- A block with only one transaction also timed out after 25 seconds.
- `address_utxo` contains approximately 345 million rows.
- The table heap is approximately 245 GB.

## Root cause

`BFBlocksStorageReaderImpl.findBlockAddressTransactions()` filters block outputs using the numeric block number:

```java
var outputsSelect = dsl.select(outputAddressField, outputUtxo.TX_HASH.as("tx_hash"))
        .from(outputUtxo)
        .where(outputUtxo.BLOCK.eq(blockNumber))
        .and(outputAddressExpr.isNotNull());
```

There is no index on the numeric `address_utxo.block` column. The Blockfrost index named `idx_address_utxo_block` actually indexes `block_hash`:

```yaml
- table: address_utxo
  indexes:
    - name: idx_address_utxo_block
      columns:
        - block_hash
```

The numeric predicate therefore produces this plan:

```text
Gather
  Workers Planned: 2
  -> Parallel Seq Scan on address_utxo
       Filter: (block = 13735342)
```

Additionally, the `affected` union is embedded once in `pagedAddresses` and again in the outer join, so the output branch can be evaluated twice. With the current numeric predicate, that can mean two scans of the large UTXO table.

The other branches are already indexed and fast:

- Spent inputs through `tx_input.spent_at_block`: 0.191 ms
- Transaction lookup through `transaction.block`: indexed

## Proposed fix

The service has already resolved a `BFBlockRow`, including both its height and hash. Pass both values to the storage method:

```java
BFBlockRow block = requireBlock(hashOrNumber);
long blockNumber = requireBlockNumber(block);

List<BFBlockAddressTxRow> rows =
        bfBlocksStorageReader.findBlockAddressTransactions(
                blockNumber,
                block.hash(),
                page,
                count
        );
```

Update the storage interface and implementation signature accordingly, then use the indexed hash for the output branch:

```java
.where(outputUtxo.BLOCK_HASH.eq(blockHash))
```

The numeric block number can remain in use for `tx_input.spent_at_block` and `transaction.block`, where matching indexes already exist.

## Validation

The hash-based output lookup used the existing `idx_address_utxo_block` index:

```text
Bitmap Heap Scan on address_utxo
  Recheck Cond: (block_hash = ...)
  -> Bitmap Index Scan on idx_address_utxo_block
       Index Cond: (block_hash = ...)
Execution Time: 1.620 ms
```

Using the hash predicate in the complete existing query shape, tested with a 57-transaction block:

```text
Execution Time: 123.932 ms
```

Materializing the `affected` union once as a CTE reduced the complete query further:

```text
Execution Time: 52.234 ms
```

The hash predicate is the essential fix. Materializing/reusing `affected` is an optional secondary optimization.

## Database-only workaround

For deployments that cannot immediately rebuild or upgrade:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_address_utxo_block_number
    ON yaci_store.address_utxo (block);

ANALYZE yaci_store.address_utxo (block);
```

This requires scanning the approximately 245 GB table and creates another multi-GB index. Reusing the existing `block_hash` index in the application is preferable.

Sotatek-HenryDo:
## Summary

`findBlockAddressTransactions()` filters `address_utxo` on the numeric `block` column, which had no index. The blockfrost index named `idx_address_utxo_block` is actually defined on `block_hash`, so the predicate degraded to a parallel sequential scan over the whole 352M-row / 244GB table.

The output/spent `unionAll` was also declared as a derived table and referenced twice (once in `paged_addresses`, again in the outer join), so Postgres inlined it at each reference and ran that sequential scan twice.

Following review, the fix is:
- **`blockfrost-index.yml`**: replace `idx_address_utxo_block` (on `block_hash`) with `idx_address_utxo_block_number` (on `block`), so the index matches the query the code has issued since #811. The unused `block_hash` index is dropped in favour of one that is actually reachable.
- **Move the union into a CTE** so it is evaluated once instead of twice.

The query itself is unchanged from `main`: no `blockHash` parameter, no signature change.

Removal of the now-obsolete `idx_address_utxo_block` on existing installations is not handled here; see discussion.

Fixes #1044

## Test plan

Verified against a synced mainnet DB (`address_utxo` 352,801,888 rows / 244GB heap, 396GB total), on the 57-transaction block 13749038, with `idx_address_utxo_block_number` built via `CREATE INDEX CONCURRENTLY`.

`EXPLAIN (ANALYZE, BUFFERS)`:
- Before: `Parallel Seq Scan on address_utxo`, `Filter: (block = ...)`, planner cost 67,736,241 — appearing **twice** in the plan (`address_utxo o` and `o_1`)
- After: `Index Scan using idx_address_utxo_block_number`, `Index Cond: (block = 13749038)`, planner cost 77,804, that node at 0.064ms
- The CTE is materialized once (`Storage: Memory, Maximum Storage: 134kB`) and scanned twice

Query time: **197ms cold, then 8.5ms / 8.5ms / 8.5ms warm.**

Row equivalence: filtering by `block = 13749038` and by the corresponding `block_hash` both return 67 rows, symmetric difference 0.

Index size on mainnet, measured after the build:
- `idx_address_utxo_block` (`block_hash`, current) — 3263 MB
- `idx_address_utxo_block_number` (`block`, new) — **2497 MB**

Endpoint timing before the fix, measured on the live mainnet node: **246.90s** for `/api/v1/blockfrost/blocks/13749038/addresses?count=100&page=1`.

### DIFF
--- a/components/dbutils/src/main/resources/blockfrost-index.yml
+++ b/components/dbutils/src/main/resources/blockfrost-index.yml
@@ -3,9 +3,9 @@
     - name: idx_address_utxo_owner_addr_full
       columns:
         - owner_addr_full
-    - name: idx_address_utxo_block
+    - name: idx_address_utxo_block_number
       columns:
-        - block_hash
+        - block
 
 - table: assets
   indexes:
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/service/BFBlockService.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/service/BFBlockService.java
@@ -117,7 +117,7 @@ public class BFBlockService {
         long blockNumber = requireBlockNumber(block);
 
         List<BFBlockAddressTxRow> rows =
-                bfBlocksStorageReader.findBlockAddressTransactions(blockNumber, block.hash(), page, count);
+                bfBlocksStorageReader.findBlockAddressTransactions(blockNumber, page, count);
 
         Map<String, LinkedHashMap<String, BFBlockAddressTxDTO>> addresses = new LinkedHashMap<>();
         for (BFBlockAddressTxRow row : rows) {
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/service/BFBlockService.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/service/BFBlockService.java
@@ -116,7 +116,8 @@ public class BFBlockService {
         BFBlockRow block = requireBlock(hashOrNumber);
         long blockNumber = requireBlockNumber(block);
 
-        List<BFBlockAddressTxRow> rows = bfBlocksStorageReader.findBlockAddressTransactions(blockNumber, page, count);
+        List<BFBlockAddressTxRow> rows =
+                bfBlocksStorageReader.findBlockAddressTransactions(blockNumber, block.hash(), page, count);
 
         Map<String, LinkedHashMap<String, BFBlockAddressTxDTO>> addresses = new LinkedHashMap<>();
         for (BFBlockAddressTxRow row : rows) {
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/BFBlocksStorageReader.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/BFBlocksStorageReader.java
@@ -27,5 +27,5 @@ public interface BFBlocksStorageReader {
 
     List<BFBlockTxCborRow> findBlockTxCbor(long blockNumber, int page, int count, Order order);
 
-    List<BFBlockAddressTxRow> findBlockAddressTransactions(long blockNumber, String blockHash, int page, int count);
+    List<BFBlockAddressTxRow> findBlockAddressTransactions(long blockNumber, int page, int count);
 }
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/BFBlocksStorageReader.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/BFBlocksStorageReader.java
@@ -27,5 +27,5 @@ public interface BFBlocksStorageReader {
 
     List<BFBlockTxCborRow> findBlockTxCbor(long blockNumber, int page, int count, Order order);
 
-    List<BFBlockAddressTxRow> findBlockAddressTransactions(long blockNumber, int page, int count);
+    List<BFBlockAddressTxRow> findBlockAddressTransactions(long blockNumber, String blockHash, int page, int count);
 }
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
@@ -189,7 +189,7 @@ public class BFBlocksStorageReaderImpl implements BFBlocksStorageReader {
     }
 
     @Override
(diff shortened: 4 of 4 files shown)


satran004:
Thanks @Sotatek-HenryDo 

@Sotatek-HuyLe3  In issue #1044  for this PR, I had mentioned using `block_hash` in the query instead of `block` to avoid adding an additional index. However, it seems the `block_hash` index has been added by `blockfrost-index.yml`, and it is not required by the core.

I am wondering if we should update the index in `blockfrost-index.yml` to use `block` number instead, and remove the `block_hash` index. We can also verify the index sizes for both. With a `block` index, we don’t need any Java code change.

huylevt:
@satran004 Good point. I checked the history again.                                                                                                                                                                               
                                                                                                                                                                                           
The original implementation in #811 queries address_utxo by block number. The block_hash index was later proposed in the gap docs in #958, and I carried it into blockfrost-index.yml in #962. It seems there is no  query that depends on this index, or any benchmark explaining why block_hash was chosen, so it looks like the index simply didn’t match the actual query.                                                              
                                                                                                                                                                                                                       
If we don’t need to handle existing installations, I agree that changing the index to block is cleaner. We can then drop the blockHash parameter change from this PR.                                                  
                                                                                                                                                                                                                   
I’d still keep the CTE change, since it avoids evaluating the affected union twice.  
                                                                                                                                                                                      
@Sotatek-HenryDo  Could you please verify?

Sotatek-HenryDo:
Thanks @satran004 @huylevt, I verified this on the mainnet DB (352M rows, 244 GB).

You're right that the `block_hash` index isn't used. `pg_stat_user_indexes` shows `idx_address_utxo_block` with only 12 scans, and those are from my own testing on this PR. Nothing in the code filters on `block_hash`, it's only written and selected as a column.

On size, both are about the same. `idx_address_utxo_slot` is a bigint index on the same table, and it's 3374 MB against 3263 MB for the current `block_hash` one, so switching wouldn't save space.

One thing to consider for existing installations: `IndexService` uses `CREATE INDEX IF NOT EXISTS`, which matches on the index name, not the definition. Installs that already applied the blockfrost indexes have `idx_address_utxo_block` on `block_hash`, so re-applying wouldn't do anything and the endpoint would stay slow. We'd need a new index name and a manual drop of the old one.

Otherwise I'm happy to switch to the `block` index if you prefer.

satran004:
Thanks @Sotatek-HenryDo . Let's create a new index `idx_address_utxo_block_number` then and remove the existing one from bf index yaml file. To remove existing index, we can add a release notes or create a new migration script in `utxo` store. So remove index if exists.

Sotatek-HenryDo:
@satran004, done, the PR now uses `idx_address_utxo_block_number` on `block` and drops the `block_hash` index from the yaml. The `blockHash` parameter change is reverted, so the query is back to what it was on `main`. The CTE change stays, as @huylevt suggested. I've updated the PR description accordingly.

I built the index on the mainnet DB with `CREATE INDEX CONCURRENTLY` and re-measured on block 13749038 (57 txs). The plan now uses `Index Scan using idx_address_utxo_block_number` instead of the parallel seq scan, and the query runs in 197ms cold, 8.5ms warm.

One correction to my earlier comment on index sizes. I had estimated from `idx_address_utxo_slot` as a bigint proxy and said it would be about the same. That was wrong, the real index is smaller than the one it replaces, 2497 MB against 3263 MB for the `block_hash` one. `slot` is much more distinct than `block`, so it doesn't dedupe nearly as well. Switching saves roughly 766 MB once the old index is dropped.

I've left the removal of the existing `idx_address_utxo_block` out of this PR. Happy to add it as a release note or a `utxo` migration, whichever you prefer.

Sotatek-HuyLe3a:
>I've left the removal of the existing idx_address_utxo_block out of this PR. Happy to add it as a release note or a utxo migration, whichever you prefer.

we might add it in release note

Sotatek-HuyLe3a (approved):

satran004:
@Sotatek-HenryDo Can you please look into the conflict?

PR #1071, opened 2026-08-07
fix(epoch-aggr): update total fees for existing epochs

### LINKED ISSUE #1064: Blockfrost API: /epochs/latest reports `fees` orders of magnitude below the real value for the in-progress epoch
### Summary

For the in-progress epoch, `GET /epochs/latest` returns a `fees` total that is roughly four orders of magnitude too small. The other aggregate fields on the same response differ only by an amount consistent with indexing lag, so this looks like a distinct aggregation bug rather than a lag artifact.

### Reproduction

Epoch 647, both queried within the same second:

| field | Blockfrost | yaci-store | consistent with lag? |
|---|---|---|---|
| `block_count` | 12313 | 12184 | yes (~1% behind) |
| `tx_count` | 52535 | 52010 | yes (~1% behind) |
| `output` | 16885019591433382 | 16599713626656832 | yes (~1.7% behind) |
| `last_block_time` | 1785954632 | 1785952265 | yes (~40 min behind) |
| **`fees`** | **16204702704** | **593236** | **no: 0.0037% of the expected value** |
| `active_stake` | 21428560878223742 | 21428560878223742 | exact match |

A 1% block deficit cannot produce a 99.996% fee deficit. The historical endpoint `/epochs/{epoch}` for closed epochs did not show this pattern in our sweep, so the suspicion is the running-total path for the current epoch rather than the stored per-epoch aggregate.

Not urgent for historical queries, but any client charting live epoch fees gets a badly wrong number.

### Environment

huylevt:
## Summary

- Persist the recalculated `totalFees` value when updating an existing epoch.
- Fix incorrect `fees` values returned for the in-progress epoch.

Fixes #1064

## Root cause

`EpochService` correctly recalculates the total fees from all blocks in an epoch. However, `EpochMapper.updateEntity()` did not copy `totalFees` to an existing `EpochEntity`.

The initial insert mapped every field, so an epoch created when its first block arrived stored that block's fee. Later aggregation runs calculated the correct accumulated fees but left the stored value unchanged.

## Changes

Update `EpochMapper.updateEntity()` to set `totalFees` together with the other recalculated epoch fields.

## Verification

- Confirmed against the mainnet database that the stored fee matched the first block's fee while the block-derived epoch total was significantly higher.
- Ran `./gradlew :aggregates:epoch-aggr:compileJava` successfully.
- Ran the `epoch-aggr` and Blockfrost epoch module checks successfully.

## Notes

- No database schema changes are required.
- The current epoch will be corrected by the next scheduled aggregation after deployment.
- Previously affected closed epochs require a separate one-time data backfill.

### DIFF
--- a/aggregates/epoch-aggr/src/main/java/com/bloxbean/cardano/yaci/store/epochaggr/storage/impl/mapper/EpochMapper.java
+++ b/aggregates/epoch-aggr/src/main/java/com/bloxbean/cardano/yaci/store/epochaggr/storage/impl/mapper/EpochMapper.java
@@ -11,6 +11,7 @@ public abstract class EpochMapper {
 
     public EpochEntity updateEntity(Epoch epoch, EpochEntity targetEntity) {
         targetEntity.setTotalOutput(epoch.getTotalOutput());
+        targetEntity.setTotalFees(epoch.getTotalFees());
         targetEntity.setTransactionCount(epoch.getTransactionCount());
         targetEntity.setBlockCount(epoch.getBlockCount());
         targetEntity.setStartTime(epoch.getStartTime());


satran004 (approved):
Thanks @huylevt . LGTM

PR #1074, opened 2026-08-07
Drop NULLS LAST from descending sorts on null-free base columns (fixes #1073)

### LINKED ISSUE #1073: Blockfrost API performance: DESC NULLS LAST ordering defeats btree indexes; /blocks/latest takes 23s and reads ~12GB per call
### Summary

`GET /blocks/latest` takes **23 seconds** on a mainnet instance and reads roughly 12 GB from disk per call. Fetching the *same block* by hash takes **3 ms**.

The cause is `ORDER BY number DESC NULLS LAST`. A default btree index is `ASC NULLS LAST`, so scanning it backwards yields `DESC NULLS FIRST`. Requesting `DESC NULLS LAST` therefore cannot use the index for ordering and PostgreSQL falls back to sorting the entire `block` table (13.7M rows) to return one row.

### Evidence

The generated query (from `pg_stat_activity` during a live request):

```sql
select b.block_time, b.number, b.hash, ...
from block as b
where true
order by b.number desc nulls last
fetch next $2 rows only
```

Same query, with and without `NULLS LAST`, on a synced mainnet instance:

| variant | plan | buffers | execution time |
|---|---|---|---|
| `order by number desc nulls last` (as shipped) | Parallel **Sort**, `Sort Key: number DESC NULLS LAST` | `read=1594145` (~12 GB) | ~1349 ms per worker, 23 s end to end via the API |
| `order by number desc` | **Index Scan Backward** using `idx_block_number` | `hit=4` | **0.039 ms** |
| `order by number desc nulls first` | **Index Only Scan Backward** using `idx_block_number` | `hit=4` | **0.049 ms** |

`select count(*) from block where number is null` returns **0**, so on mainnet the `NULLS LAST` qualifier changes nothing semantically while costing a full-table sort.

Endpoint-level, same instance, quiet box:

```
GET /blocks/latest          23.4 s
GET /blocks/{same hash}      0.003 s
GET /blocks/{any number}     0.003 s
```

### Scope

This is not limited to one query. `.desc().nullsLast()` appears **16 times** across the Blockfrost extension:

| module | occurrences |
|---|---|
| `asset` | 8 |
| `block` | 6 |
| `scripts` | 2 |

(`asc().nullsLast()` appears 10 more times and is harmless: it matches the default index ordering.)

Every `desc().nullsLast()` on a column whose index is a plain btree will have this shape. That likely contributes to other slow endpoints we reported, e.g. #1066 and #1067, both of which order descending over large tables.

### Suggested fix

For columns that are never null in practice, dropping `.nullsLast()` on the descending sorts restores index usage (`DESC` alone gives `NULLS FIRST`, which the backward index scan satisfies). `block.number` is the clearest case: zero nulls, and `/blocks/latest` is one of the most frequently called endpoints in any Blockfrost client, typically on a health-check or polling loop.

Where a column genuinely can be null and `NULLS LAST` semantics matter, the alternative is an index declared with matching ordering, e.g. `CREATE INDEX ... ON block (number DESC NULLS LAST)`, at the cost of an extra index per sort.

Happy to send a PR for whichever shape you prefer. I can also re-measure any of the 16 sites on this mainnet instance if that helps prioritise.

### Environment

Beta3 mainnet validation instance (Gero/Nexus). Build: `main` + #847 + #865 + #866, blockfrost + ledger-state profiles, synced to tip, full `blockfrost-index` optional set applied, PostgreSQL 18.4, 16 vCPU / 64 GB / pd-ssd. Measurements taken on an otherwise idle box, serially.

adamcazes:
Fixes #1073.

## Problem

`ORDER BY <col> DESC NULLS LAST` cannot be satisfied by a backward scan of a default btree index (which yields `DESC NULLS FIRST`), so PostgreSQL sorts the entire table. Measured on a synced mainnet instance: `GET /blocks/latest` takes **~23s** and reads **~12GB** of buffers per call; the identical query with plain `DESC` is an Index Scan Backward at **0.039ms**. Full EXPLAIN evidence on #1073.

## Scope: 9 of the 16 sites, each justified by mainnet null counts

| file | sites | column | nulls on mainnet |
|---|---|---|---|
| `BFBlocksStorageReaderImpl` | 4 | `block.number` | 0 of 13.7M |
| `BFAssetStorageReaderImpl` | 1 | `assets.slot` | 0 of 19M |
| `BFAssetStorageReaderImpl` | 2 | `transaction.block` | 0 of 122.6M |
| `BFScriptsStorageReaderImpl` | 1 | `script.slot` | 0 of 152k |
| `BFScriptsStorageReaderImpl` | 1 | `transaction_scripts.slot` | 0 of 59.9M |

**Deliberately unchanged:**
- The 4 `transaction.tx_index` sorts: that column has **14,505 null rows** on mainnet, so `NULLS LAST` is semantically load-bearing; those sorts are also secondary keys inside block-filtered sets where the qualifier is harmless.
- The 3 derived-field sorts (a `min()` aggregate and subquery-table fields): they order materialized results, so there is no base-table index to defeat.
- All 10 `asc().nullsLast()` sites: `ASC NULLS LAST` is the default btree ordering, no cost.

## Effect

With this change `/blocks/latest` on our mainnet instance goes from ~23s to sub-millisecond query time (endpoint round trip a few ms). The asset and scripts list endpoints get the same treatment for their primary descending sorts; happy to post before/after measurements for those too once this is rebuilt into our validation image.

🤖 Assisted by Claude Code; every change compile-checked and every null count measured against mainnet data.

### DIFF
--- a/extensions/blockfrost/asset/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/asset/storage/impl/BFAssetStorageReaderImpl.java
+++ b/extensions/blockfrost/asset/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/asset/storage/impl/BFAssetStorageReaderImpl.java
@@ -172,7 +172,7 @@ public class BFAssetStorageReaderImpl implements BFAssetStorageReader {
     public List<BFAssetHistory> findAssetHistory(String unit, int page, int count, Order order) {
         int offset = Math.max(page, 0) * count;
         Condition assetCondition = buildAssetCondition(unit);
-        SortField<?> slotOrder = order == Order.desc ? ASSETS.SLOT.desc().nullsLast() : ASSETS.SLOT.asc().nullsLast();
+        SortField<?> slotOrder = order == Order.desc ? ASSETS.SLOT.desc() : ASSETS.SLOT.asc().nullsLast();
         SortField<?> txOrder = order == Order.desc ? ASSETS.TX_HASH.desc() : ASSETS.TX_HASH.asc();
 
         var query = dsl.select(ASSETS.TX_HASH, ASSETS.MINT_TYPE, ASSETS.QUANTITY)
@@ -205,7 +205,7 @@ public class BFAssetStorageReaderImpl implements BFAssetStorageReader {
         Field<String> txHashField = TRANSACTION.TX_HASH;
         Field<Integer> txIndexSortField = TRANSACTION.TX_INDEX;
         SortField<?> blockOrder = order == Order.desc
-                ? TRANSACTION.BLOCK.desc().nullsLast()
+                ? TRANSACTION.BLOCK.desc()
                 : TRANSACTION.BLOCK.asc().nullsLast();
         SortField<?> txIndexOrder = order == Order.desc
                 ? txIndexSortField.desc().nullsLast()
@@ -240,7 +240,7 @@ public class BFAssetStorageReaderImpl implements BFAssetStorageReader {
         Field<Long> txIndexField = DSL.coalesce(txIndexSortField, 0).cast(Long.class).as("tx_index");
 
         SortField<?> blockOrder = order == Order.desc
-                ? TRANSACTION.BLOCK.desc().nullsLast()
+                ? TRANSACTION.BLOCK.desc()
                 : TRANSACTION.BLOCK.asc().nullsLast();
         SortField<?> txIndexOrder = order == Order.desc
                 ? txIndexSortField.desc().nullsLast()
--- a/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
+++ b/extensions/blockfrost/block/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/block/storage/impl/BFBlocksStorageReaderImpl.java
@@ -76,7 +76,7 @@ public class BFBlocksStorageReaderImpl implements BFBlocksStorageReader {
         var confirmationsField = confirmationsField(block);
 
         return blockBaseSelect(block, nextBlockField, confirmationsField, block.SLOT.eq(slot))
-                .orderBy(block.NUMBER.desc().nullsLast())
+                .orderBy(block.NUMBER.desc())
                 .limit(1)
                 .fetchOptional(record -> toBlockRow(record, block, nextBlockField, confirmationsField));
     }
@@ -93,7 +93,7 @@ public class BFBlocksStorageReaderImpl implements BFBlocksStorageReader {
                 confirmationsField,
                 block.EPOCH.eq(epoch).and(block.EPOCH_SLOT.eq(epochSlot))
         )
-                .orderBy(block.NUMBER.desc().nullsLast())
+                .orderBy(block.NUMBER.desc())
                 .limit(1)
                 .fetchOptional(record -> toBlockRow(record, block, nextBlockField, confirmationsField));
     }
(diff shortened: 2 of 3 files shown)


satran004:
Partially fixed in PR https://github.com/bloxbean/yaci-store/pull/1046

adamcazes:
Rebased onto current main. You're right that #1046 already fixed `findLatestBlock` (and its `isNotNull() + desc()` shape is a nice belt-and-suspenders variant); that site is now dropped from this PR and main's version stands.

What remains here are the **8 sites #1046 did not cover**, same pattern, same evidence: the other block-list orderings (`/blocks` pagination and `/blocks/{hash}/next|previous` walk the same `NUMBER.desc().nullsLast()` full sort), the asset module's `ASSETS.SLOT` and `TRANSACTION.BLOCK` primary sorts, and both scripts-module slot sorts. All on columns with zero nulls on mainnet (counts in the PR body). The nullable `tx_index` sorts and the derived-field sorts stay untouched as before.

adamcazes:
New evidence that raises this PR's stakes: **it fixes two Blockfrost SDK methods that currently time out.**

We ran the full `@blockfrost/blockfrost-js` read surface (84 method invocations, real mainnet identifiers, 12s timeout) against a synced mainnet instance on current `main`. Two of the seven timeouts are the sites this PR changes:

| SDK method | endpoint | on current main | sibling for contrast |
|---|---|---|---|
| `blocksPrevious()` | `/blocks/{hash}/previous` | **timeout >12s** | `blocksNext()` 29 ms |
| `epochsPrevious()` | `/epochs/{n}/previous` | **timeout >12s** | `epochsNext()` 112 ms |

The next/previous asymmetry is the tell, and the query confirms it. `findPreviousBlocks` runs:

```java
blockBaseSelect(block, nextBlockField, confirmationsField, block.NUMBER.lt(blockNumber))
        .orderBy(block.NUMBER.desc().nullsLast())
```

Measured on the instance, same predicate, same limit/offset:

```
order by number desc nulls last   ->  Sort (Sort Key: number DESC NULLS LAST)   47,925 ms
order by number desc              ->  Index Scan Backward using idx_block_number      2.1 ms
```

**~48 seconds to ~2 ms**, which is exactly the change in this PR. `epochsPrevious` follows the same pattern.

So the 8 remaining sites here are not a tidy-up: at least two of them are the difference between an SDK method working and hanging. Happy to add a regression test if you'd like one, and to re-run the full SDK matrix against a build with this merged so the before/after is on record.

satran004 (approved):
Thanks @adamcazes . LGTM

@huylevt

huylevt (approved):

adamcazes:
Post-merge verification on mainnet, and a correction to what we claimed in this thread.

We said this PR fixes two timing-out SDK methods, `blocksPrevious` and `epochsPrevious`. Rebuilt today from `bf1d09c2b` (this PR merged, plus #1070) and re-ran the SDK matrix:

| method | before | after |
|---|---:|---:|
| `blocksPrevious` | 12008 ms (timeout) | 27 ms |
| `epochsPrevious` | 12007 ms (timeout) | 12007 ms (timeout) |

`blocksPrevious` is confirmed fixed, 27 ms in the SDK matrix and 23 ms measured directly at `count=100`.

`epochsPrevious` is refuted, and our attribution was simply wrong. This PR changed the asset, block and scripts readers; it never touched the epoch module, so there was no mechanism by which it could have helped that endpoint. We should have checked the file list against the claim before making it.

The real cost in `/epochs/{n}/previous` is the `active_stake` enrichment, which sums `epoch_stake` for every epoch in the requested page: 0.949 s at `count=1`, 3.654 s at `count=5`, 13.583 s at `count=20`, linear at roughly 0.68 s per epoch, projecting to about 68 s at the SDK default `count=100`. At the DB level a single epoch's `sum(amount)` scans 1,347,175 rows in about 1000 ms.

Filed separately as #1089 with the code path and a suggested direction, since it is unrelated to the `NULLS LAST` work here.

Sorry for the noise in this thread. The `blocksPrevious` number stands on its own, and the ordering fix does what it says.

PR #1076, opened 2026-08-07
Add tx_input(spent_tx_hash) to blockfrost-index.yml (fixes #1075)

### LINKED ISSUE #1075: Blockfrost API performance: /txs/{hash} takes ~10s; tx_input has no index leading with spent_tx_hash
### Summary

`GET /txs/{tx_hash}` takes **~10 seconds** on mainnet for every transaction, regardless of era or size. Hosted Blockfrost serves the same request in ~50 ms. `GET /txs/{tx_hash}/utxos` is also affected (~9-10s).

### Root cause

The transaction detail query computes eight correlated `count(*)` sub-selects. Timed individually on a quiet, fully indexed mainnet instance:

| subquery | time | plan |
|---|---|---|
| `withdrawal.tx_hash` | 0.07 ms | Index Only Scan |
| `delegation.tx_hash` | 0.10 ms | Index Only Scan |
| `stake_registration.tx_hash` | 0.07 ms | Index Only Scan |
| `pool_registration.tx_hash` | 0.08 ms | Index Only Scan |
| `pool_retirement.tx_hash` | 0.08 ms | Index Only Scan |
| `transaction_scripts.tx_hash` | 0.10 ms | Index Only Scan |
| `assets.tx_hash` | 0.08 ms | Index Only Scan |
| **`tx_input.spent_tx_hash`** | **9723.62 ms** | Index Only Scan over the wrong index |

No index on `tx_input` leads with `spent_tx_hash`. The closest, `idx_tx_input_spent_at_block_spent_tx_hash (spent_at_block, spent_tx_hash)`, has the needed column second, so the "index only" scan still walks a huge range. `tx_input` is ~131 GB / ~300M rows on mainnet.

### Verified fix

```sql
CREATE INDEX CONCURRENTLY idx_tx_input_spent_tx_hash ON yaci_store.tx_input (spent_tx_hash);
```

Measured on this instance: build **14m21s** (concurrent, box serving), size **13 GB**. Effect:

| | before | after |
|---|---|---|
| the subquery | 9723 ms | **1.4 ms** |
| `GET /txs/{hash}` (three txs, eras Shelley through current) | ~10 s each | **0.16-0.20 s** |
| `GET /txs/{hash}/utxos` | ~9.9 s | **0.009-0.015 s** |

PR adding it to `blockfrost-index.yml` follows.

### Environment

Beta3 mainnet validation instance (Gero/Nexus): `main` + #847/#865/#866, synced to tip, full `blockfrost-index` set applied, PostgreSQL 18.4, 16 vCPU / 64 GB / pd-ssd. Serial measurements on an idle box.

adamcazes:
Fixes #1075.

One-entry addition to `blockfrost-index.yml`: an index on `tx_input (spent_tx_hash)`.

`GET /txs/{tx_hash}` computes `count(*) from tx_input where spent_tx_hash = ?` among its eight correlated sub-selects; the other seven are sub-millisecond, this one is **9.7s** on mainnet because no index leads with `spent_tx_hash` (the composite `(spent_at_block, spent_tx_hash)` has it second).

Verified on our mainnet validation instance before opening this PR:

| | before | after |
|---|---|---|
| the subquery | 9723 ms | 1.4 ms |
| `GET /txs/{hash}` | ~10 s | 0.16-0.20 s |
| `GET /txs/{hash}/utxos` | ~9.9 s | 0.009-0.015 s |
| build cost | | 14m21s concurrent, 13 GB |

Note: this file is also touched by #1060 (different table block); the hunks don't overlap, whichever merges second rebases trivially.

### DIFF
--- a/components/dbutils/src/main/resources/blockfrost-index.yml
+++ b/components/dbutils/src/main/resources/blockfrost-index.yml
@@ -39,6 +39,12 @@
     - name: idx_tx_input_tx_hash
       columns:
         - tx_hash
+    # /txs/{hash} computes count(*) where spent_tx_hash = ?; without a
+    # leading spent_tx_hash index that is a ~10s range walk on mainnet
+    # (see issue #1075). 14m21s to build, 13 GB, endpoint drops to ~0.2s.
+    - name: idx_tx_input_spent_tx_hash
+      columns:
+        - spent_tx_hash
 
 - table: block
   indexes:


satran004:
@adamcazes This seems to have already been fixed through a query change in https://github.com/bloxbean/yaci-store/pull/1050. Can you update from the main branch and try again?

adamcazes:
Confirmed, you're right: rebuilt the instance from current main (including #1050) and **dropped the index entirely** so the query change stands alone:

```
/txs/{hash}        : 0.757s cold, then 0.030-0.036s warm   (was ~10s pre-#1050)
/txs/{hash}/utxos  : 0.041s                                 (was ~9.9s)
```

That's not just sufficient, it's better than this PR's index bolted onto the old query shape (0.16-0.20s in my earlier measurement). #1050's CTE rewrite removed the need for the index rather than working around it. Withdrawing this PR; also dropped the index from our instance, which returns its 13 GB.

Bonus data point from the same rebuild: `/blocks/latest` on current main is now 0.004-0.032s (#1046 confirmed at mainnet scale too).

PR #1079, opened 2026-08-09
fix(blockfrost): Align Blockfrost network era history boundaries

### LINKED ISSUE #1063: Blockfrost API: /network/eras leaves the current era's `end` null instead of projecting it
### Summary

The last entry of `GET /network/eras` has `end: null` in yaci-store. Blockfrost projects the current era's end from the known epoch length and safe zone.

The module comes from the pending network PR #866.

### Reproduction

```jsonc
// Blockfrost, last era
{"start":{"time":219024000,"slot":133660800,"epoch":507},
 "end":{"time":279936000,"slot":194572800,"epoch":648},
 "parameters":{"epoch_length":432000,"slot_length":1,"safe_zone":129600}}

// yaci-store, last era
{"start":{"time":219024000,"slot":133660800,"epoch":507},
 "end":null,
 "parameters":{"epoch_length":432000,"slot_length":1,"safe_zone":129600}}
```

All earlier eras match exactly. Clients that use `/network/eras` to convert slots to wall-clock time for the current era get a null boundary and have to special-case it. The projection is computable from data already in the response (`start` plus `epoch_length` times the epochs elapsed), so this may just be an unimplemented branch for the open-ended era.

### Environment

Sotatek-HuyLe3a:
cc @adamcazes 

## Summary

Fixes the Blockfrost-compatible `GET /api/v1/blockfrost/network/eras` endpoint so the active era includes a projected `end` boundary instead of returning `null`, and networks that skipped eras still expose the canonical Cardano era history expected by Blockfrost.

The active boundary follows the current chain tip and the era safe zone. Missing historical eras are represented as zero-length summaries at the next observed era boundary. Both behaviors match the live Blockfrost API across mainnet, preprod, and preview.

Fixes #1063.

This PR targets `bf_network_apis`, the source branch of #866.

## Problem

Historical eras have a following era in storage, so their end epoch, slot, and time can be derived from the next era's start boundary.

The active era has no following era record. The existing implementation therefore left its `end` value unset, while Blockfrost returns a finite projected boundary for the active era.

The existing implementation also returned only era transitions observed in storage. Preview started directly in Alonzo at slot zero, so Shelley, Allegra, and Mary were absent from the response. Blockfrost preserves the canonical Hard Fork Combinator (HFC) history by returning these skipped eras as zero-length summaries.

## Changes

- Read the latest indexed block through `BlockStorageReader` when projecting the active era boundary.
- Project the end epoch using the current epoch slot and the Shelley safe zone:

```java
endEpoch = currentEpoch
        + (epochSlot >= epochLength - safeZone ? 2 : 1);
```

- Convert the projected epoch boundary to the corresponding absolute slot and time through `EraService`.
- Build an in-memory canonical HFC timeline through the latest observed era.
- Reuse the next observed boundary for a missing historical era, producing the zero-length summaries expected for Preview without modifying stored era data or exposing future eras.
- Return HTTP `503 Service Unavailable` if the latest block information required for the projection is unavailable.
- Add the explicit `stores:blocks` dependency to the Blockfrost network module.

## Safe-zone behavior

The safe zone is the period after the current ledger tip in which an unknown era transition is guaranteed not to occur.

If the tip plus the safe zone remains within the current epoch, the projected boundary is the start of the next epoch. If it reaches or crosses that boundary, the next epoch is still inside the guaranteed safe zone, so the projection moves to the start of the following epoch.

### Blockfrost compatibility

Compared a local read-only Yaci Store instance backed by each network database with the corresponding live Blockfrost API. The full normalized JSON payloads matched:

| Network | Active-era boundary | Full `/network/eras` payload |
|---|---|---|
| Mainnet | Exact match | Exact match, 7 eras |
| Preprod | Exact match | Exact match, 7 eras |
| Preview | Exact match | Exact match, 7 eras including 3 zero-length placeholders |

Preview starts directly in Alonzo. The response now preserves the canonical era order by emitting zero-length Shelley, Allegra, and Mary summaries at slot, epoch, and time zero before the actual Alonzo summary.

## Database impact

- No schema migration
- No data migration
- No database writes


satran004:
@huylevt Looks good to me. We need to change the following method in BFNetworkService to return YS host/port.

```
  public BFRootDto getRoot() {
        return BFRootDto.builder()
                .url("https://cardano-mainnet.blockfrost.io/api")
                .version(BF_API_VERSION)
                .build();
    }
```

adamcazes:
@satran004 the `getRoot()` change is now on the #866 branch: commit 61faf4b97 makes it read `blockfrost.hostname` + `blockfrost.apiPrefix` from the environment instead of the hardcoded hosted URL, so a self-hosted deployment reports its own origin. It was sitting unpushed on our side from the earlier review round, apologies for the lag. The branch update also merges your conditional-properties fix and main; compile is clean. #1079 should pick it up on its next rebase against the branch.

On #1079 itself: happy to validate the era boundaries on our synced mainnet instance once it lands on the branch, same as we did for #1059. We filed #1063, so we have the before-state captured for comparison.

Sotatek-HuyLe3a:
Thanks @adamcazes for the update. Since #1079 targets bf_network_apis, I’ll merge it directly into the branch, so the resulting branch will include both 61faf4b and the era-boundary changes.                                  
                                                                                                                                                                                                                Once it lands, we’d appreciate your help validating the era boundaries on the synced mainnet instance, using the before-state from #1063 for comparison.

adamcazes:
We validated the era boundaries on our mainnet validation instance.

Environment: yaci-store built from `bf_network_apis` tip 923e586b4 merged into our combined build (upstream `main` at bf1d09c2b, plus the pools, governance and network extension modules). Mainnet sync at tip, blockfrost and ledger-state profiles, PostgreSQL 18.4, 16 vCPU / 64 GB / pd-ssd. At measurement time the instance was at epoch 648, epoch slot 307269, tip block 13794980, tip age 3.0 seconds.

The before column is the build immediately preceding this one on the same instance and the same database, which is the state we reported in #1063. The after column is this branch. Blockfrost was queried at the same time.

Era starts, time / slot / epoch:

| era | before | after | Blockfrost | after == Blockfrost |
|---|---|---|---|---|
| 0 | 0 / 0 / 0 | 0 / 0 / 0 | 0 / 0 / 0 | yes |
| 1 | 89856000 / 4492800 / 208 | 89856000 / 4492800 / 208 | 89856000 / 4492800 / 208 | yes |
| 2 | 101952000 / 16588800 / 236 | 101952000 / 16588800 / 236 | 101952000 / 16588800 / 236 | yes |
| 3 | 108432000 / 23068800 / 251 | 108432000 / 23068800 / 251 | 108432000 / 23068800 / 251 | yes |
| 4 | 125280000 / 39916800 / 290 | 125280000 / 39916800 / 290 | 125280000 / 39916800 / 290 | yes |
| 5 | 157680000 / 72316800 / 365 | 157680000 / 72316800 / 365 | 157680000 / 72316800 / 365 | yes |
| 6 | 219024000 / 133660800 / 507 | 219024000 / 133660800 / 507 | 219024000 / 133660800 / 507 | yes |

Era ends, time / slot / epoch:

| era | before | after | Blockfrost | after == Blockfrost |
|---|---|---|---|---|
| 0 | 89856000 / 4492800 / 208 | 89856000 / 4492800 / 208 | 89856000 / 4492800 / 208 | yes |
| 1 | 101952000 / 16588800 / 236 | 101952000 / 16588800 / 236 | 101952000 / 16588800 / 236 | yes |
| 2 | 108432000 / 23068800 / 251 | 108432000 / 23068800 / 251 | 108432000 / 23068800 / 251 | yes |
| 3 | 125280000 / 39916800 / 290 | 125280000 / 39916800 / 290 | 125280000 / 39916800 / 290 | yes |
| 4 | 157680000 / 72316800 / 365 | 157680000 / 72316800 / 365 | 157680000 / 72316800 / 365 | yes |
| 5 | 219024000 / 133660800 / 507 | 219024000 / 133660800 / 507 | 219024000 / 133660800 / 507 | yes |
| 6 | null | 280800000 / 195436800 / 650 | 280800000 / 195436800 / 650 | yes |

Current era end: before `null`, after `280800000 / 195436800 / 650`, Blockfrost `280800000 / 195436800 / 650`.

Difference between our projected current era end and Blockfrost's: 0 seconds, 0 slots, 0 epochs.

The two responses compare equal as full JSON documents, including the `parameters` object of every era. 14 of 14 boundaries match. One boundary changed between before and after, the era index 6 end. The other 13 were unchanged.

On the projection horizon: epoch slot at measurement was 307269, epoch length 432000, safe zone 129600. 307269 >= 432000 - 129600, so the projection resolves to 648 + 2 = 650, and Blockfrost returned that same boundary. We measured at one point in the epoch, inside the safe zone, so this run does not separate the +1 and +2 branches. We can re-run outside the safe zone if that is useful to you.

Timing, GET /network/eras, 5 consecutive runs: 6.3ms, 4.9ms, 4.9ms, 5.0ms, 5.2ms.

Regression check, GET /genesis: HTTP 200 in 10.0ms, response equal to Blockfrost's.

Thanks for the fix.

PR #1087, opened 2026-08-11
feat: Add analytics-query module (DuckDB query layer over Parquet)

satran004:
## Summary

Adds `aggregates:analytics-query`, a DuckDB-based read layer over the data exported by `analytics-store` (DuckLake catalog + Parquet files), and makes `analytics-store` **write-only** — every read path (schema discovery, pre-built statistics, ad-hoc SQL) now lives in `analytics-query`, which the MCP server (#1088) will build on.

**Query layer**
- `ParquetReadConnectionProvider` — one sandboxed in-memory DuckDB instance; readers are `DuckDBConnection.duplicate()` handles behind a semaphore (no HikariCP pool of separate instances). Views are built over the **committed** DuckLake data files (via `DuckLakeCatalogSnapshotReader`, which borrows the writer's connection — DuckDB allows one instance per catalog file per process) and refreshed every 5 minutes.
- Optional live PostgreSQL federation (`UnifiedViewBuilder`, `CutoffSlotResolver`, `postgres_scanner`) — **off by default**. Unified views `UNION ALL` the exported rows (`boundary <= cutoff`) with live rows (`boundary > cutoff`); the cutoff is the last contiguous completed partition (slot for DAILY tables, epoch for EPOCH tables). Exporters declare the metadata (`TableExporter.getFederationBoundaryColumn()`, `getSourceColumnMappings()`), so tables like `tx_input` (`spent_at_slot`), `reward` (`earned_epoch`), `epoch` (`number`) federate correctly; 36 of 41 tables verified row-for-row against PostgreSQL.
- `AnalyticsQueryExecutor` — per-query timeout, semaphore concurrency, sanitized errors, and **bounded results**: every statement runs as `SELECT * FROM (<sql>) AS q LIMIT n+1` (the Superset/Metabase/DuckDB-UI approach). Ad-hoc SQL defaults to 100 rows (`query.default-max-rows`), request `maxRows` up to the hard limit (`query.max-rows`, 10,000); truncation is reported via `X-Analytics-Row-Limit` / `X-Analytics-Truncated`.
- `SqlValidator` — single SELECT/WITH statement; token-based, quote/comment aware; blocks DDL/DML, file/URL access, extension management, catalog/metadata functions and the attached PostgreSQL alias.
- `AnalyticsSchemaService` — table/column discovery with cached row counts, `dataScope` (`historical` / `historical+live`) and query hints.
- `AnalyticsQueryController` under **`/api/v1/analytics/query/*`** (`/schema`, `/schema/{table}`, `/blocks/*`, `/transactions/*`, `POST /sql`), with OpenAPI descriptions; the legacy read endpoints in `analytics-store` are removed.

**analytics-store**
- Write path only (`DuckDbConnectionHelper`, exporters, admin API under `/api/v1/analytics/admin/*`); reader pool removed.
- **Export finality guard** (`17588773b`): daily partitions are exported only once every row is older than the network's rollback bound. The window is derived from genesis — `k × slotLength / activeSlotsCoeff` (2160 × 20 s = 12 h on mainnet/preprod, 2.4 h on preview) — plus `continuous-sync.finality-margin-hours` (1); the export end date is the UTC date of (latest block time − window) minus `buffer-days`. Previously `buffer-days=1` meant 0–24 h behind the tip (the youngest rows of a day could be minutes old at export time, and the pipeline has no rollback handling); now a UTC day is exported ~13 h after it ends. `buffer-days` default 2 → 1; `finality-window-hours` override; `< 1` / negative values clamped with a warning; `GET /admin/status` reports `finalityWindowHours`. The dead `finalization-lag-days` property and the uncalled `UniversalExportService.exportDailyTables()/exportEpochTables()` are removed.
- **DuckDB JDBC 1.4.4 → 1.5.5.** DuckLake catalog format 0.3 → 1.0 with `AUTOMATIC_MIGRATION` on the writer attach (one-way — back up `ducklake.catalog.db` before upgrading an existing deployment); data inlining disabled/flushed so rows are always visible in Parquet.
- Precise "already attached" / file-handle-conflict detection, credential redaction in attach/query errors.

**Cross-cutting**
- Swagger split into three groups (`core-apis`, `analytics`, `blockfrost`) in `applications/all`.
- Docs: new *Analytics Query API* page plus updates to the analytics overview/configuration/querying pages.

Gated behind `yaci.store.analytics.query.enabled` (default `false`); the REST endpoints additionally require `yaci.store.analytics.query.rest-api-enabled=true` (unauthenticated — opt in deliberately).

## Review status

The items flagged in the original description (validator file-read bypasses, `OFFSET`/`ASSET` false positives, no tests) and every follow-up review round are **fixed in this branch** (`70c2d5a7d`, `412a9c6ca`, `13421d936`, `c8173ba5a`, `469e4692c`; `f8a92776c`/`17588773b` add the export-finality guard): engine-level sandbox (`allowed_directories`, `enable_external_access=false`, `lock_configuration`), token-based validator, and 100+ tests across the two modules.

The last round (`469e4692c`) covered the row-limit change: the query timeout is now enforced for the whole execution (no JDBC result streaming — with it, `setQueryTimeout` only covered the execute phase), truncation headers are CORS-exposed, malformed request bodies return the same `{"error"}` 400 shape, `/schema` advertises the effective row limits, and docs/OpenAPI wording matches the behaviour.

Follow-ups for the MCP server are tracked for #1088 (it must switch from the removed `isTruncated()` to `execute(sql, getHardRowLimit())`, and will likely want a per-call query timeout above the 30 s REST default).

## Test plan

- `JAVA_HOME=<jdk21> ./gradlew :aggregates:analytics-store:test :aggregates:analytics-query:test :applications:all:build` — passes (analytics-store 86 tests, analytics-query 35 tests).
- Verified on a synced preprod node (federation on and off): all `/api/v1/analytics/query/*` and `/api/v1/analytics/admin/*` endpoints, ad-hoc SQL (limits, headers, CORS, validator/DuckDB/malformed-body error responses, 30 s timeout observed on heavy queries), unified views reconciled against PostgreSQL, DuckLake catalog migration to 1.5.5 on a copy first; export-finality guard verified via startup log / `admin/status` (`exportEndDate` moved from tip−1 day to the finalized-tip rule; override path exercised, then restored) and an independent review of the window arithmetic.


aggregates/analytics-query/src/main/java/com/bloxbean/cardano/yaci/store/analytics/query/service/SqlValidator.java
```diff
@@ -0,0 +1,481 @@
...
+                                "File path or URL references are not allowed in queries");
+                    }
+                    i = end + delimiter.length() - 1;
+                }
+            }
+            i++;
+        }
+    }
+
+    private static boolean isPathLikeLiteral(String literal) {
```
Sotatek-HuyLe3a: I think the SQL validator may still allow DuckDB replacement scans through ordinary relative path literals.
  `isPathLikeLiteral()` currently blocks absolute paths, `./`, `../`, URLs, etc., but it does not seem to block paths such as:
  ```sql
  SELECT * FROM 'data/analytics/some-file.parquet'
  SELECT * FROM 'block/date=2026-01-01/file.parquet'
  SELECT * FROM 'some.csv'
```

satran004:
## Review follow-up update

Pushed commit [`624b50339`](https://github.com/bloxbean/yaci-store/commit/624b5033944b43bf5d06ae061bcade1074a9ad60) with the fixes discussed during review.

Here is the update in simple terms:

- **Safer SQL filter:** blocks DuckDB file replacement scans through bare or quoted file names, blocks dollar quoting, serialized JSON SQL execution, Parquet metadata functions, and `repeat()` resource abuse. The filter remains deliberately small and conservative so it is easier to audit and debug.
- **Correct live federation range:** exported data is used only inside the actual contiguous exported range. PostgreSQL supplies rows before and after that range, so partial/admin backfills no longer hide older PostgreSQL history.
- **Correct exporter mappings:** added the missing `instant_reward` mapping and aligned `unclaimed_reward_rest` with the other reward tables.
- **Empty DuckLake tables:** committed tables with a schema but no files are now exposed as safe zero-row views instead of disappearing.
- **Safer DuckLake snapshot reads:** snapshot discovery coordinates with the writer and keeps the previous committed snapshot while a write is active.
- **Better table metadata:** fixed partition metadata, documented optional tables and their fallbacks, and clarified REST versus MCP balance guidance.
- **More accurate schema information:** `dataAsOf` now comes from completed daily exports instead of the wall clock. Successful schema metadata is refreshed every 30 minutes; failed scans retry every 5 minutes. A failed `DESCRIBE` now returns 503 instead of a misleading 200 with no columns.
- **Clear error responses:** invalid SQL remains 400, real query timeouts return 408, and full query capacity returns 503. Timeout detection now uses JDBC timeout types/SQL state rather than matching words in error messages.
- **Configuration cleanup:** the writer now respects `duckdb.threads`, dead Hikari settings were removed, and memory/admin documentation was corrected.
- **Tests and docs:** added regression coverage for exported ranges, empty DuckLake tables, metadata, mappings, writer settings, SQL bypasses, and real timeout behavior. Query-layer documentation was updated as well.

Validation status:

- ✅ `commit-build` passed on JDK 21
- ✅ `Test deployment` passed
- ⚠️ SonarCloud quality gate is still red with C security/reliability ratings and needs a separate follow-up

The dependent MCP PR #1088 has also been updated with this parent branch.

Sotatek-HuyLe3a (approved):

Sotatek-HuyLe3a:
Look good to me ! @satran004

Issue #1067, opened 2026-08-05
Blockfrost API performance: /metadata/txs/labels takes ~7-10s on mainnet


adamcazes:
### Summary

`GET /metadata/txs/labels` takes 7 to 10 seconds on mainnet against ~65 ms on hosted Blockfrost.

### Reproduction

```
yaci-store /metadata/txs/labels?count=100&page=1 : 7.2s   (10.5s in an earlier run)
Blockfrost /metadata/txs/labels?count=100&page=1 : 0.065s
```

Full `blockfrost-index` optional index set applied. The endpoint returns per-label counts over the whole `transaction_metadata` table (57 GB class on mainnet), so the shape looks like a full aggregation per request; a materialized count or a periodically refreshed summary would probably be more appropriate than an index.

Side note: the returned counts themselves differ from Blockfrost by 1 to 2 per label (e.g. `4741795` vs `4741794`), which is consistent with indexing lag rather than a counting error.

### Environment

Found during the beta3 mainnet validation run (Gero/Nexus real-world beta testing).
Build: `main` + #847 + #865 + #866, blockfrost + ledger-state profiles, full mainnet sync
(tip epoch 647, AdaPot caught up), PostgreSQL 18.4, 16 vCPU / 64 GB / pd-ssd, repo Docker
stack with shipped compose defaults. Compared against hosted Blockfrost with
`scripts/compare-blockfrost` at the same moment; every example below was re-verified
by hand afterwards with direct requests to both APIs.

huylevt:
Verified on another mainnet instance: the endpoint took 3.36–3.57s across three runs.                                                                                                                                  
transaction_metadata is 62 GB (~137.4M rows), and EXPLAIN ANALYZE took 4.83s.                                                                                                                                          
PostgreSQL used the existing idx_txn_metadata_label, but still scanned ~137.5M index entries before aggregation and LIMIT.

adamcazes:
Useful that yours is 3.36-3.57s where ours was 7.2-10.5s, because the gap is probably a PostgreSQL config difference rather than data volume, and it feeds the "small to high config recommendations" idea @satran004 raised in the thread.

[...]

PR #1091, opened 2026-08-11
Fix(blockfrost): drep delegators performance

### LINKED ISSUE #1081: Blockfrost API performance: `/governance/dreps/{drep_id}/delegators` takes 1.7–2.3s on mainnet
## Summary

`GET /governance/dreps/{drep_id}/delegators` takes **1.7–2.3 seconds** on a fully synced mainnet instance, an order of magnitude above every other governance route measured in the same run.

This issue is split out from [#1067](https://github.com/bloxbean/yaci-store/issues/1067). All benchmark results below were produced and reported by [`@adamcazes`](https://github.com/adamcazes) in [his PostgreSQL configuration follow-up test](https://github.com/bloxbean/yaci-store/issues/1067#issuecomment-5230982044).

The endpoint remained slow under both the shipped compose defaults and the maintainer deployment configuration, suggesting that the query shape, rather than PostgreSQL memory configuration, is the primary bottleneck.

## Measurements

According to `@adamcazes`, the same endpoint set was run serially on the same instance, with the same data and build, while the machine was otherwise idle. Each result is the median of five samples after discarding a cold first request.

| PostgreSQL configuration | Median latency |
|---|---:|
| Shipped compose defaults (`shared_buffers=128MB`) | **2270 ms** |
| Maintainer configuration (`shared_buffers=4GB`) | **1711 ms** |

The maintainer configuration improves this endpoint by approximately 25%, but it remains substantially slower than the other governance routes from the same test:

| Endpoint | Observed latency |
|---|---:|
| `/governance/dreps` | 11–13 ms |
| `/governance/dreps/{id}` | 27–35 ms |
| `/governance/dreps/{id}/updates` | 5 ms |
| `/governance/dreps/{id}/votes` | 5–6 ms |
| `/governance/proposals` | 42–64 ms |
| `/governance/proposals/{tx}/{ix}` | 5 ms |
| `/governance/proposals/{tx}/{ix}/votes` | 6–7 ms |
| **`/governance/dreps/{id}/delegators`** | **1711–2270 ms** |

Source: [`@adamcazes`' test results in #1067](https://github.com/bloxbean/yaci-store/issues/1067#issuecomment-5230982044).

huylevt:
## Summary

Optimizes the Blockfrost-compatible endpoint:

```http
GET /api/v1/blockfrost/governance/dreps/{drep_id}/delegators
```

The endpoint now resolves and paginates current delegators before loading balances, uses `stake_address_balance` instead of reconstructing UTxOs from full history, and calculates Blockfrost-compatible controlled amounts from an AdaPot checkpoint plus subsequent reward changes.

The change also aligns stake-address and DRep registration lifecycle handling with ledger behavior, including deregistration, re-registration, DRep unregistration, and DRep re-registration.

Closes #1081.

Related: #1067, #865.

## Problem

The previous implementation performed expensive work before pagination:

1. Rank the full `delegation_vote` history to resolve the latest delegation per address.
2. Join delegators to historical `address_utxo` rows.
3. Anti-join those outputs with `tx_input` to remove spent outputs.
4. Aggregate balances for all matching delegators.
5. Apply `LIMIT` and `OFFSET` only after the balance reconstruction.

On the mainnet database used for verification, `address_utxo` contained approximately 353 million rows and `tx_input` approximately 342 million rows. A request for only 100 delegators still read about 413,778 historical outputs, more than 99% of which were later discarded as spent.

The old implementation also had correctness gaps:

- Ordering only by `slot` could select the wrong certificate when several transactions or certificates shared a slot.
- Stake deregistration did not clear historical vote delegations from the result.
- Re-registering a stake address could incorrectly revive a delegation from its previous registration cycle.
- DRep unregistration and re-registration cycles were not considered.
- The returned `amount` represented only the current UTxO balance, while Blockfrost returns the controlled amount, including rewards not yet withdrawn.

### Resolve and paginate current delegators first

- Select only delegations whose stake address is currently registered.
- Require the delegation to belong to the stake address's current registration cycle.
- Require the target DRep to be currently registered and the delegation to belong to its current registration cycle.
- Treat only `REG_DREP_CERT` and `UNREG_DREP_CERT` as DRep lifecycle boundaries; `UPDATE_DREP_CERT` does not start a new cycle.
- Resolve chain order with `(slot, tx_index, cert_index)`.
- Exclude a delegation when the same address has a newer delegation to any DRep.
- Apply ordering and pagination before any balance lookup, bounding the remaining work to at most the requested page size.

### Use current balance snapshots

The endpoint now reads the latest `stake_address_balance.quantity` for each paged delegator instead of rebuilding current UTxOs from `address_utxo` and `tx_input` history.

### Calculate the amount reported by Blockfrost

Blockfrost reports the delegator's controlled amount, not only the ADA currently held by its UTxOs:

```text
amount = current UTxO balance + rewards not yet withdrawn
```

Replaying the complete reward history for every address would make the endpoint slow. Instead, the endpoint starts from the latest completed AdaPot stake snapshot and applies only the changes that are not already represented by that snapshot.

If the latest completed `REWARD_CALC` job is for epoch `N`, AdaPot has materialized `epoch_stake` for epoch `E = N - 1`. For each delegator on the requested page, the calculation is:

1. Read `epoch_stake(E).amount`, which already contains the address's UTxO balance and rewards not yet withdrawn at the checkpoint.
2. Subtract the address's UTxO balance at `E` to recover the reward balance already included in the checkpoint.
3. Add rewards not represented by the checkpoint and subtract withdrawals made after it.
4. Add the reward balance remaining after withdrawals to the current UTxO balance.

```text
reward_at_checkpoint = epoch_stake(E).amount - utxo_balance_at_E

current_reward = max(
    reward_at_checkpoint
    + rewards_not_in_checkpoint
    - withdrawals_not_in_checkpoint,
    0
)

response.amount = current_utxo_balance + current_reward
```

For example:

```text
epoch_stake(E).amount       = 1,100 ADA
utxo_balance_at_E           = 1,000 ADA
reward_at_checkpoint        =   100 ADA
later rewards               =    20 ADA
later withdrawals           =    30 ADA
rewards not yet withdrawn   =    90 ADA
current UTxO balance        = 1,250 ADA
response.amount             = 1,340 ADA
```

If the address has no `epoch_stake(E)` row, the endpoint cannot use the checkpoint. It therefore calculates the reward balance from the address's complete reward, reward-rest, instant-reward, and withdrawal history. This fallback preserves correctness for addresses omitted from the snapshot, such as a still-active DRep delegator whose stake pool has retired.

<details>
<summary>Reward boundaries used after checkpoint E</summary>

The delta query mirrors the rules used to build `epoch_stake(E)` so that a reward is neither skipped nor counted twice:

| Change type | Included in the delta when |
|---|---|
| Member or leader reward | `spendable_epoch > E` and either `earned_epoch > E` or `spendable_epoch > E + 1` |
| Refund | `spendable_epoch > E` |
| Reward rest | `spendable_epoch > E` |
| Instant reward | `spendable_epoch > E + 1` |
| Withdrawal | `epoch > E`, subtracted from the reward balance |

The static `spendable_epoch` lower bounds also allow PostgreSQL to prune reward partitions that are already represented by the checkpoint.

</details>

### Module dependencies

The governance extension now declares direct dependencies on the account, AdaPot, staking, and transaction modules that own the tables used by the query. 

No database index, schema migration, or stored-data migration is required.

## Ledger lifecycle behavior

The implementation follows the Conway ledger behavior in which:

- Stake credential deregistration clears its vote delegation.
- Re-registering the stake credential does not restore a delegation from an earlier cycle.
- `ConwayUnRegDRep` clears delegations to that DRep.
- Re-registering the DRep starts with an empty delegation set; delegators must delegate again.
- Updating a registered DRep does not start a new lifecycle.

As a regression case, a retired DRep whose historical delegation rows previously selected two delegators now returns HTTP 200 with `[]`, matching Blockfrost.

## Performance verification

Verification used a local read-only deployment connected to the configured PostgreSQL mainnet database.

Request used for the corrected old-versus-new comparison:

```http
GET /api/v1/blockfrost/governance/dreps/drep1y29h2q6cst2pvkl2sqqvf5ljcy36uv7pmy482xnczddzgqshus24w/delegators?page=1&count=100&order=asc
```

| Implementation | Result | Latency |
|---|---|---:|
| Old `issue-864` implementation | Client timeout without response | >180,098.3 ms |
| Fixed branch, warm median | HTTP 200, 100 rows | 69.2 ms |
| Fixed branch, warm maximum | HTTP 200, 100 rows | 138.2 ms |
| Blockfrost during the corrected old-version run | HTTP 200, 100 rows | 830.7 ms |

The measured lower bound is more than 1,300 times faster than the old implementation for the same DRep, page, count, order, and mainnet database configuration.

Across six warm pagination/order cases, local medians ranged from 32 ms to 141 ms. Adam's original PostgreSQL measurements in #1067 were 1,711-2,270 ms, providing a separate environment-level baseline.

## Blockfrost comparison

At a synchronized chain tip, six combinations of `page`, `count`, and `order` matched Blockfrost exactly for delegator addresses and ordering.

The checkpoint calculation was also compared with a full-history calculation for 100 addresses:

```text
matched addresses: 100/100
max_abs_delta:      0
sum_abs_delta:      0
```

A later mainnet fallback-focused sample used checkpoint epoch `647`:

```text
active delegators:              79,418
addresses without checkpoint:    3,156
observed fallback rate:          3.9739%
```

For a DRep with 59 fallback addresses out of 115 local delegators, all 59 addresses were present on Blockfrost and all 59 amounts matched exactly. One example without `epoch_stake(647)` produced `4152` lovelace from both local Yaci Store and Blockfrost.

## Follow-up issues

The fallback-focused comparison identified two issues that are not caused by the reward-checkpoint calculation and should be investigated separately:

- **AdaPot active-stake/reward parity:** For all 22 checkpoint-backed addresses whose amounts differed, the checkpoint result matched Yaci Store's full-history calculation exactly. The entire `12,148` lovelace difference came from member rewards earned in epoch `646`. Yaci Store's total active stake for that epoch was `156,993,730,649` lovelace higher than Blockfrost's, which produced slightly higher pool and member rewards. A follow-up should determine why the AdaPot `epoch_stake` snapshot begins to diverge at epoch `646`; changing the DRep checkpoint query would only hide the upstream data difference.
- **Conway bootstrap delegation lifecycle:** The additional Blockfrost delegator delegated to the DRep immediately before that DRep's registration certificate in the same transaction. The Conway ledger permits this ordering during the bootstrap phase, but the current query requires the delegation certificate to occur at or after the DRep registration certificate. A follow-up should use the latest DRep unregistration as the delegation-cycle boundary, while still requiring the DRep's current lifecycle state to be registered, so re-registration cannot revive delegations cleared by an earlier unregistration.

These findings do not change the performance result or the verified checkpoint-versus-full-history parity. They limit claims of full Blockfrost result parity until the upstream AdaPot snapshot difference and the bootstrap-era lifecycle edge case are resolved.

## Verification

The following commands completed successfully with Java 21:

```bash
./gradlew :extensions:blockfrost:blockfrost-governance:test
./gradlew :applications:all:bootJar
```

Focused storage tests cover pagination, full chain ordering, redelegation, stake registration cycles, DRep registration cycles, reward checkpoint boundaries, full-history fallback, and negative reward clamping. The focused test source is intentionally not included in this PR yet; it remains outside the committed diff for separate review.

At the current remote head (`3e77a7bb1`), both SonarCloud checks pass. No separate GitHub build/test check is currently reported for this draft PR.

## Known trade-offs

- Full-history fallback favors correctness over latency. It represented approximately 3.97% of active delegators in the measured mainnet snapshot, but can be concentrated within individual DReps. A page containing 59 fallback addresses took approximately 7.06 seconds on its first local request.
- The response is assembled from multiple read statements without a shared repeatable-read snapshot. Consolidating the reads or adding a common transaction snapshot is outside this PR's current scope.
- Focused test source is not yet part of the PR, although it was executed locally.

### DIFF
--- a/extensions/blockfrost/governance/build.gradle
+++ b/extensions/blockfrost/governance/build.gradle
@@ -6,6 +6,7 @@ dependencies {
     api project(":stores:governance")
     api project(":stores-api:governance-api")
     api project(":aggregates:governance-aggr")
+    api project(":aggregates:account")
     api project(":stores:utxo")
 
     implementation 'org.springframework.boot:spring-boot-starter-web'
--- a/extensions/blockfrost/governance/build.gradle
+++ b/extensions/blockfrost/governance/build.gradle
@@ -7,10 +7,15 @@ dependencies {
     api project(":stores-api:governance-api")
     api project(":aggregates:governance-aggr")
     api project(":aggregates:account")
+    api project(":aggregates:adapot")
+    api project(":stores:staking")
+    api project(":stores:transaction")
     api project(":stores:utxo")
 
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.jooq:jooq'
+
+    testImplementation 'com.h2database:h2'
 }
 
 publishing {
--- a/extensions/blockfrost/governance/build.gradle
+++ b/extensions/blockfrost/governance/build.gradle
@@ -15,7 +15,6 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.jooq:jooq'
 
-    testImplementation 'com.h2database:h2'
 }
 
 publishing {
--- a/extensions/blockfrost/governance/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/governance/storage/impl/BFGovernanceStorageReaderImpl.java
+++ b/extensions/blockfrost/governance/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/governance/storage/impl/BFGovernanceStorageReaderImpl.java
@@ -2,7 +2,6 @@ package com.bloxbean.cardano.yaci.store.blockfrost.governance.storage.impl;
 
 import com.bloxbean.cardano.client.crypto.Bech32;
 import com.bloxbean.cardano.client.util.HexUtil;
-import com.bloxbean.cardano.yaci.store.blockfrost.common.util.BlockfrostDialectUtil;
 import com.bloxbean.cardano.yaci.store.blockfrost.governance.storage.BFGovernanceStorageReader;
 import com.bloxbean.cardano.yaci.store.blockfrost.governance.storage.impl.model.BFDRepDelegator;
 import com.bloxbean.cardano.yaci.store.blockfrost.governance.storage.impl.model.BFDRep;
@@ -18,6 +17,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.jooq.DSLContext;
 import org.jooq.JSON;
+import org.jooq.Select;
 import org.jooq.SortField;
 import org.jooq.SortOrder;
 import org.jooq.impl.DSL;
@@ -30,12 +30,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static com.bloxbean.cardano.yaci.store.account.jooq.Tables.STAKE_ADDRESS_BALANCE;
 import static com.bloxbean.cardano.yaci.store.epoch.jooq.Tables.EPOCH_PARAM;
 import static com.bloxbean.cardano.yaci.store.governance.jooq.Tables.*;
 import static com.bloxbean.cardano.yaci.store.governance_aggr.jooq.Tables.DREP_DIST;
 import static com.bloxbean.cardano.yaci.store.governance_aggr.jooq.Tables.GOV_ACTION_PROPOSAL_STATUS;
-import static com.bloxbean.cardano.yaci.store.utxo.jooq.Tables.ADDRESS_UTXO;
-import static com.bloxbean.cardano.yaci.store.utxo.jooq.Tables.TX_INPUT;
 
 @Slf4j
 @Component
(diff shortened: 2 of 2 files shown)


huylevt:
>AdaPot active-stake/reward parity: For all 22 checkpoint-backed addresses whose amounts differed, the checkpoint result matched Yaci Store's full-history calculation exactly. The entire 12,148 lovelace difference came from member rewards earned in epoch 646. Yaci Store's total active stake for that epoch was 156,993,730,649 lovelace higher than Blockfrost's, which produced slightly higher pool and member rewards. A follow-up should determine why the AdaPot epoch_stake snapshot begins to diverge at epoch 646; changing the DRep checkpoint query would only hide the upstream data difference.

 Resolved after repairing the epoch-transition rollback handling (https://github.com/bloxbean/yaci-store/pull/1093) and replaying mainnet data from epoch 644.

satran004 (approved):

PR #1099, opened 2026-08-13
[2.0.x]  fix(governance): backport SPO default votes per proposal 

huylevt:
Back-port: #1034 
Excluded the integration test.


satran004 (approved):
LGTM. Thanks @huylevt

Issue #1084, opened 2026-08-10
Blockfrost API: /accounts/{stake_address}/registrations omits the documented `deposit` field


adamcazes:
### Summary

`GET /accounts/{stake_address}/registrations` returns entries without the `deposit` key. Blockfrost documents the field and returns it on every entry.

From the Blockfrost OpenAPI spec, schema `account_registration_content`, the `deposit` property is type `string`, nullable, described as:

> Deposit in Lovelaces paid at this registration. `null` on `deregistered` rows.

We found this with a field-level response comparison run with `@blockfrost/blockfrost-js` against both backends for the same stake addresses: the hosted Blockfrost response carries `deposit`, the yaci-store response omits the key entirely. Everything else on the endpoint matched.

### Why the field is absent

`stake_registration` has no `deposit` column. Live schema on our synced mainnet instance:

```
tx_hash, cert_index, tx_index, credential, cred_type, type, address,
epoch, slot, block_hash, block, block_time, update_datetime
```

So this is a store-level gap surfaced through the Blockfrost mapper, not a mapper omission that can be fixed in the extension alone.

The store side is already tracked in #451 and #896, which both ask for the deposit amount to be kept on `stake_registration`, and #896 enumerates the Conway certificate types that carry it. This issue records the Blockfrost API-surface consequence, filed on its own per the per-gap convention with the `data_gap` label.

### Domain note

Conway-era `reg_cert` and the combined `stake_reg_deleg_cert` family carry an explicit deposit coin in the certificate itself. Pre-Conway registrations carry no coin: the amount is the protocol parameter `keyDeposit` in force at registration time. Two workable shapes:

1. Store the certificate deposit at ingest, and backfill the pre-Conway rows as null from the epoch protocol parameters. This matches what #451 and #896 already propose and keeps the read path cheap.
2. Join epoch protocol parameters at read time for rows with no stored deposit.

One detail either way: Blockfrost returns `null` for `deregistered` entries, so deregistration rows should map to `null` rather than to the refund amount.

### Related

Same family as #1062, where `/governance/dreps/{drep_id}/updates` omits `deposit` for the same reason on the governance side. One systematic decision about where certificate deposits are stored and how they are read back would likely close both endpoints together.

### Environment

[...]

Issue #451, opened 2025-02-20
Track stake registration deposit in stake_registration table


satran004:
Currently, the stake registration deposit is hard-coded as 2 ADA. To calculate the total stake deposit, 2 ADA is multiplied by the number of registered stake keys.

With governance enabled, the stake key deposit amount can now be changed by a governance vote. Therefore, it's better to keep the stake key deposit amount during stake key registration.

PR #1101, opened 2026-08-18
fix(blockfrost): return stake registration deposit from current epoch params

Sotatek-HenryDo:
Addresses #1084.

@adamcazes — this is the workaround discussed in #1084.

## Problem

`GET /accounts/{stake_address}/registrations` omits the `deposit` field that Blockfrost documents and returns on every entry.

It cannot be served from storage today: `stake_registration` has no `deposit` column, and `StakeRegProcessor` drops the certificate's `getCoin()` at ingest.

## Change

Blockfrost-layer workaround, per the direction in #1084. Storing the per-certificate deposit is tracked separately in #451.

| file | change |
|---|---|
| `BFAccountRegistrationDto` | add `deposit` (lovelace string) |
| `BFAccountMapper` | `@Mapping(target = "deposit", ignore = true)` — owned by the service |
| `BFAccountService.findRegistrations` | resolve the current epoch's `keyDeposit` once per call; set on `registered`, `null` on `deregistered` |
| `BFAccountServiceTest` | new — 3 cases |

`keyDeposit` comes from `EpochParamClient.getLatestProtocolParams()`, injected as an `ObjectProvider` so the endpoint still answers when the epoch store is disabled — `deposit` is `null` there instead of the request failing.

No schema change, no processor change, no new module dependency (`components:client` already arrives via `stores:staking`).

## Verification

Read-only against synced instances:

| network | epoch | case | `deposit` |
|---|---|---|---|
| mainnet | 649 | Conway registration | `"2000000"` |
| mainnet | 649 | deregistration | `null` |
| mainnet | 649 | pre-Conway registration (block 4,490,613) | `"2000000"` |
| preprod | 307 | Conway registration | `"2000000"` |
| preprod | 307 | deregistration | `null` |
| preprod | 307 | pre-Conway registration (block 48) | `"2000000"` |

Pre-Conway rows are covered because the value comes from epoch parameters, not the certificate.

New tests: key deposit available → set on registrations only; protocol params absent → `null`; epoch store disabled → `null`.

## Notes

It reads `keyDeposit` from the current epoch, as originally specified. Switching to the registration's own epoch would need a new `EpochParamClient` method, so it is left to #451.

The genesis-file fallback for scoped indexers is not included — #451 supersedes it.

### DIFF
--- a/extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
+++ b/extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
@@ -53,7 +53,7 @@ public class BFAccountService {
                 .stream()
                 .map(registration -> {
                     var dto = mapper.toRegistrationDto(registration);
-                    if ("registered".equals(dto.getAction()))
+                    if (BFAccountMapper.ACTION_REGISTERED.equals(dto.getAction()))
                         dto.setDeposit(keyDeposit);
                     return dto;
                 })
--- a/extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
+++ b/extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
@@ -3,9 +3,12 @@ package com.bloxbean.cardano.yaci.store.blockfrost.account.service;
 import com.bloxbean.cardano.yaci.store.blockfrost.account.dto.*;
 import com.bloxbean.cardano.yaci.store.blockfrost.account.mapper.BFAccountMapper;
 import com.bloxbean.cardano.yaci.store.blockfrost.account.storage.BFAccountStorageReader;
+import com.bloxbean.cardano.client.api.model.ProtocolParams;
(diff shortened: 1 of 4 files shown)


extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
```diff
@@ -43,8 +47,27 @@ public List<BFAccountDelegationDto> findDelegations(String stakeAddress, int pag
...
 
     public List<BFAccountRegistrationDto> findRegistrations(String stakeAddress, int page, int count, Order order) {
+        String keyDeposit = currentKeyDeposit();
+
         return storageReader.findRegistrations(stakeAddress, page, count, order)
-                .stream().map(mapper::toRegistrationDto).toList();
+                .stream()
+                .map(registration -> {
+                    var dto = mapper.toRegistrationDto(registration);
+                    if ("registered".equals(dto.getAction()))
```
satran004: @Sotatek-HenryDo Can we use a constant for "registered" ? May be create in BFAccountMapper.java or separate BFAccountConstants and use. There are also other similar text "deregistered", "refund"

Sotatek-HenryDo: Updated

Sotatek-HuyLe3a (approved):

PR #1102, opened 2026-08-18
fix(assets-ext): propagate CIP-26 token deletions from the upstream registry

matiwinnetou:
## Problem

The CIP-26 sync only ever added or updated token metadata. Tokens whose mapping file was **removed** from the upstream GitHub token registry were never deleted locally, so yaci-store kept serving stale metadata forever — in both incremental and full sync.

## Solution

Backport of the deletion handling from cf-token-metadata-registry (`fix/cip26-sync-deletions`), adapted to the assets-ext design:

- **`GitService.getChangedFiles` → `getChangedMappings`**, returning a new `ChangedMappings` record with both upserted files and deleted mapping file names from the tree diff between the last synced commit and HEAD.
- **Incremental sync**: deleted mapping file names (minus `.json`) become subjects to delete. For canonical registry entries the filename equals the subject; for mismatched spam files that were never indexed, the delete is a harmless no-op.
- **Full sync reconciliation**: DB subjects with no corresponding mapping file in the registry are deleted — covers removals that happened while commit-hash tracking was unavailable. An empty mappings folder is treated as a broken clone rather than "everything was deleted" and skips cleanup as a safety measure.
- **`Cip26MetadataService.deleteMapping`** classifies failures via a new `DeleteOutcome` enum, mirroring the existing `InsertOutcome` semantics: transient failures block the `last_commit_hash` advance so the next sync retries; permanent failures are logged loudly and skipped past so the sync can't loop forever.

## Design divergences from the cf original

- cf's `deleteMapping` returns a boolean and any failure blocks the hash advance; here failures are classified transient vs. permanent, consistent with this codebase's insert path.
- cf deletes from two tables (logo has an FK on metadata); here logos live on the same `cip26_metadata` row, so deletion is a single `deleteById`.

No schema change — no DB migration needed.

## Testing

- Ported cf's `GitServiceTest` deletion cases (incl. delete-then-re-add within a range reads as a modify, non-JSON deletions filtered out).
- New sync-service cases: incremental deletion, transient delete failure blocks the cursor, permanent delete failure advances it, full-sync stale-subject cleanup, empty-mappings-folder safety.
- New `deleteMapping` / `findAllSubjects` service tests.
- `./gradlew :extensions:assets-ext:test` — all green (JDK 21).

### DIFF
--- a/extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip26/model/ChangedMappings.java
+++ b/extensions/assets-ext/src/main/java/com/bloxbean/cardano/yaci/store/extensions/assetstore/cip26/model/ChangedMappings.java
@@ -1,0 +1,19 @@
+package com.bloxbean.cardano.yaci.store.extensions.assetstore.cip26.model;
+
+import java.nio.file.Path;
+import java.util.List;
+
+/**
+ * Result of diffing the mappings folder between two commits of the token registry repository.
+ *
+ * @param upsertedFiles    mapping files that were added or modified and must be (re-)indexed
+ * @param deletedFileNames file names (e.g. {@code <subject>.json}) of mapping files that were
+ *                         removed from the registry and whose metadata must be deleted locally
+ */
+public record ChangedMappings(List<Path> upsertedFiles, List<String> deletedFileNames) {
+
+    public static ChangedMappings empty() {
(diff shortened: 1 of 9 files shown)


Sotatek-HuyLe3a (approved):

satran004:
@matiwinnetou @Sotatek-HuyLe3a  Need to cherry-pick for release/2.0.x branch

PR #1105, opened 2026-08-19
fix(governance): correct voting procedure indexes

### LINKED ISSUE #1104: Governance voting procedure index is derived with incorrect scope
This issue was found while resolving Blockfrost governance compatibility gaps in [PR #865](https://github.com/bloxbean/yaci-store/pull/865).

## Summary

This PR fixes the derived voting procedure position used by governance vote responses so it matches the ordering expected by Blockfrost-compatible APIs.

The Conway ledger stores voting procedures as:

```text
Map Voter (Map GovActionId VotingProcedure)
```

That means the action position is scoped inside each voter's governance-action map. It is not a single sequence shared by every vote in a transaction or block event.

## What Is Blockfrost `cert_index`?

Blockfrost exposes governance votes with a `cert_index` field for compatibility with its transaction certificate-style response shape.

For Conway voting procedures, this value is not a ledger certificate index and is not stored directly by the Cardano ledger. It is a derived position inside one voter's ordered governance-action map for a transaction. In practice, the position starts at `0` for each voter in each transaction.

Because the position is voter-scoped, `(tx_hash, cert_index)` is not unique for governance votes. A transaction with multiple voters can legitimately have several vote rows whose Blockfrost `cert_index` is `0`.

## Root cause

Yaci Store uses one incrementing index across the entire `GovernanceEvent`.
cardano-db-sync derives the index independently for each voter's governance-action map with [`zip [0 ..] actions`](https://github.com/IntersectMBO/cardano-db-sync/blob/eb8e862de8f4076523eb4146905bab7eac1d5e1e/cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/GovAction.hs#L293-L327).
Consequently, Yaci Store returns incorrect `cert_index` values for multi-voter or multi-action votes.

The index resets for every voter and every transaction; `(tx_hash, cert_index)` is therefore not unique.

## Mainnet examples verified with db-sync

The following Yaci values were observed before applying this fix. Blockfrost matches db-sync in both examples.

| Mainnet scenario | db-sync | Yaci Store |
|---|---|---|
| Multi-voter transaction `0dc918e1345eb17e1f2a0240aa78bfde149be4a768aa7c4095bd351e6955d85b` | DRep: `0, 1, 2, 3`; SPO: `0` | DRep: `0, 1, 2, 3`; SPO: `4` |

## Proposed fix

1. Derive the index separately inside each voter loop for each transaction.
2. Preserve an explicit inner `GovActionId` ordering compatible with ledger/db-sync semantics.
3. Keep the public JSON field name `cert_index` for Blockfrost compatibility.

## Verification

Add regression coverage for:

- Multiple governance actions from one voter: `0, 1, 2, ...`.
- Multiple voters in one transaction: each voter starts at `0`.
- Multiple transactions in one event: each transaction starts over.

Historical db-sync data confirms these cases, including a transaction with 28 distinct SPO voters that all have index `0`.

huylevt:
Fixes #1104 
## Summary

This PR corrects the index assigned to governance voting procedures.

- Reset the index for each voter's governance-action map in each transaction.
- Order governance actions by `(transactionId, actionIndex)` before deriving the index, matching `GovActionId` ordering in cardano-ledger.
- Add regression coverage for multiple actions, voters, and transactions.

## Context

The issue was discovered while resolving Blockfrost governance compatibility gaps in [#865](https://github.com/bloxbean/yaci-store/pull/865), but the root cause is in the governance store ingestion logic rather than the Blockfrost extension.

Cardano ledger represents voting procedures as `Map Voter (Map GovActionId VotingProcedure)`. The previous processor flattened all transactions and voters with one event-wide counter and relied on Java `Map` iteration order. This could produce incorrect and non-deterministic indexes for multi-voter or multi-action transactions.

## Verification performed

- [x] `./gradlew :stores:governance:test --tests com.bloxbean.cardano.yaci.store.governance.processor.VotingProcedureProcessorTest`
- [x] `./gradlew :stores:governance:test`
- [x] Regression test verifies ordering by governance action ID and index reset across voters and transactions.

## Required preprod verification

- [x] Sync a clean Yaci Store instance again on preprod, or re-index governance data from before the selected voting transactions. Restarting an instance with existing rows is not sufficient.
- [ ] Verify a transaction containing multiple governance actions from one voter receives contiguous indexes `0, 1, 2, ...` in `(transactionId, actionIndex)` order.
- [ ] Verify a transaction containing multiple voters starts the index again at `0` for each voter.
- [ ] Verify multiple voting transactions processed in the same event do not carry indexes across transaction boundaries.
- [ ] At an aligned chain tip, compare the affected proposal-vote endpoints with Blockfrost preprod and confirm the returned `cert_index` values match.

## Data compatibility

There is no schema or public API change. Existing voting procedure rows are not rewritten automatically; historical verification requires a fresh sync, re-index, or a separately designed backfill.

## Follow-up issue

  `voting_procedure.idx` is also used by `VotingAggrService#getVotesByDRep` to compare a vote with `drep_registration.cert_index` when both belong to the same transaction.

  These indexes represent positions in different transaction-body collections and therefore are not directly comparable. The ledger processes `CERTS` before `GOV` and removes votes associated with DReps unregistered in the
  transaction without comparing certificate and voting-procedure indexes.

 A follow-up issue should remove this cross-collection index comparison and add coverage for same-transaction DRep unregistration and voting.

### DIFF
--- a/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/VotingProcedureProcessor.java
+++ b/stores/governance/src/main/java/com/bloxbean/cardano/yaci/store/governance/processor/VotingProcedureProcessor.java
@@ -18,6 +18,7 @@ import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -30,6 +31,10 @@ import static com.bloxbean.cardano.yaci.store.governance.GovernanceStoreConfigur
 @Slf4j
 public class VotingProcedureProcessor {
 
+    private static final Comparator<GovActionId> GOV_ACTION_ID_ORDER = Comparator
+            .comparing(GovActionId::getTransactionId)
+            .thenComparingInt(GovActionId::getGov_action_index);
+
     private final VotingProcedureStorage votingProcedureStorage;
     private final ApplicationEventPublisher publisher;
 
@@ -39,7 +44,6 @@ public class VotingProcedureProcessor {
         EventMetadata eventMetadata = governanceEvent.getMetadata();
 
         List<VotingProcedure> votingProcedures = new ArrayList<>();
-        int index = 0;
 
         for (TxGovernance txGovernance : governanceEvent.getTxGovernanceList()) {
             if (txGovernance.getVotingProcedures() == null) {
@@ -53,8 +57,14 @@ public class VotingProcedureProcessor {
             for (var entry : voting.entrySet()) {
                 Voter voter = entry.getKey();
                 var votingInfoMap = entry.getValue();
+                int index = 0;
 
-                for (var votingInfoEntry : votingInfoMap.entrySet()) {
+                // Ledger GovActionId ordering compares transactionId first, then actionIndex.
+                var orderedVotingInfo = votingInfoMap.entrySet().stream()
+                        .sorted(Map.Entry.comparingByKey(GOV_ACTION_ID_ORDER))
+                        .toList();
+
+                for (var votingInfoEntry : orderedVotingInfo) {
                     VotingProcedure votingProcedure = new VotingProcedure();
 
                     var govActionId = votingInfoEntry.getKey();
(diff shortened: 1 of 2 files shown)


satran004 (approved):
Thanks @huylevt LGTM

PR #1107, opened 2026-08-20
feat(transaction): store full transaction CBOR behind save-full-tx-cbor

### LINKED ISSUE #1090: Blockfrost API: /txs/{hash}/cbor returns the transaction body, not the full transaction
### Summary

`GET /txs/{hash}/cbor` returns the CBOR of the transaction body only. Blockfrost returns the full serialized transaction, the 4 element array of body, witness set, `is_valid` and auxiliary data.

Measured on the same newly synced transaction against both backends:

| | yaci-store | hosted Blockfrost |
|---|---:|---:|
| payload | 453 bytes | 1327 bytes |
| leading byte | `0xa9` | `0x84` |

`0xa9` is a CBOR map of 9 pairs, which is the transaction body. `0x84` is a CBOR array of 4 elements, which is the full transaction. yaci-store's payload is a strict prefix of Blockfrost's once the array header is skipped, so the body bytes themselves agree exactly. What is absent is the witness set, the `is_valid` flag and the auxiliary data.

The `size` field on `/txs/{hash}` comes from the same stored bytes and differs the same way: 453 from yaci-store against 1326 from Blockfrost for that transaction. Blockfrost reports 1326 for a 1327 byte payload; we have not chased that one byte.

### Root cause

`TransactionProcessor.collectTransactionCbor` (`stores/transaction/.../processor/TransactionProcessor.java`, lines 272 to 291) persists the body CBOR:

```java
String cborHex = transaction.getBody().getCbor();
```

That is the only `getCbor()` call in the transaction store, and there is no option to store the full transaction instead. `transaction_cbor.cbor_data` therefore holds body bytes and `cbor_size` is the body length, which both the `/cbor` endpoint and the `size` field inherit.

### Consequence for consumers

Any client that deserializes the response as a transaction fails, since a Blockfrost-shaped consumer expects the 4 element array the API contract implies. A body-only payload does not round-trip through a standard transaction decoder, cannot be checked against its witnesses, and cannot be resubmitted. It is usable for inspecting the body and not much beyond that.

Related: #711, which originally asked for tx CBOR storage, describes wanting the transaction CBOR as seen at the N2N protocol level for pre-signature vetting. That use case also needs the full transaction rather than the body.

### Suggested direction

A Blockfrost-equivalent response needs the full serialized transaction. The witness set, `is_valid` flag and auxiliary data all arrive in the same block payload the processor is already handling, so the full transaction could be assembled at ingest, or the upstream reader could expose a full-transaction CBOR the way it currently exposes the body CBOR.

Two decisions travel with it: whether to keep the body-only form as an option given the storage difference (about 3x on this sample), and whether `size` should move to the full length at the same time, since the two need to stay consistent.

### Relationship to #1086

#1086 covers the configuration side: `store.cardano.return-tx-body-cbor` and `store.transaction.save-cbor` both have to be set before anything is stored at all. Those two flags are necessary but not sufficient for `/txs` parity. They fix the 404 and the null `size`; this issue is what remains after that.

### Environment

Mainnet validation instance (Gero/Nexus), rebuilt today from current `main` at `bf1d09c2b` plus the three pending module branches, both CBOR flags enabled, synced to tip, PostgreSQL 18.4, 16 vCPU / 64 GB / pd-ssd. Compared against our hosted Blockfrost project for the same transaction hash at the same moment.

Happy to supply both payloads for the transaction in question, and to help implement whichever shape you prefer.

Sotatek-HenryDo:
## Summary

Adds `store.transaction.save-full-tx-cbor` to store the full signed transaction CBOR (body, witness set, `isValid`, auxiliary data) in `transaction_cbor.cbor_data` instead of body-only CBOR. Fixes #1090.

Builds on #1002 and replaces its local reassembly with Yaci's byte-exact splicing:

- merge #1002 (property wiring, processor changes, fallback)
- bump Yaci to 0.4.5
- set `YaciConfig.returnFullTxCbor` when the property is on, persist `Transaction.getTxCbor()`
- drop `FullTxCborReassembler` and its test

## Why not reassemble locally

#1002 rebuilt the transaction from parsed model objects via cardano-client-lib, which can't guarantee byte equality with the on-chain transaction. Witness map ordering, native scripts round-tripping through JSON, and auxiliary data losing its original shape all produce bytes that differ from what was submitted while still decoding to an equivalent transaction — same class of issue as bloxbean/yaci#37.

Only Yaci holds the original block bytes. bloxbean/yaci#169 (in 0.4.5) splices the exact body, witness-set and auxiliary-data segments per the era CDDL and exposes them as `Transaction.getTxCbor()`. No encoder touches the bytes, so the result is byte-exact by construction and returns `null` rather than wrong CBOR.

## Behavior

| `save-cbor` | `save-full-tx-cbor` | stored |
|---|---|---|
| `false` | `false` | nothing |
| `true` | `false` | body-only CBOR (unchanged) |
| any | `true` | full tx CBOR, falling back to body-only when Yaci returns `null` |

`save-full-tx-cbor` works standalone. Both flags default to `false`, so existing deployments are unaffected.

`cbor_size` derives from the stored bytes, so it grows to the full length with the payload — keeps `/txs/{hash}/cbor` and `size` on `/txs/{hash}` consistent, one of the open questions in #1090.

## Verification

Unit tests cover the standalone case and the fallback when `txCbor` is absent.

Synced preprod and mainnet windows with the flag on, then checked every payload: era-correct array shape, `blake2b-256` of element 0 equal to `tx_hash`, no trailing bytes. A sample per window was byte-compared against Koios `tx_cbor`.

| Window | Txs | Shape | Body hash | vs Koios |
|---|---:|---|---|---|
| Preprod tip | 287 | `array(4)` | 287/287 | 280/280 identical |
| Mainnet tip (Conway) | 710 | `array(4)` | 710/710 | 300/300 identical |
| Mainnet epoch 400 (Babbage) | 42219 | `array(4)` | 42219/42219 | 400/400 identical |
| Mainnet epoch 270 (Mary) | 15834 | `array(3)` | 15834/15834 | 300/300 identical |

59,050 transactions checked, 1,280 byte-compared, no mismatches. Mary covers the pre-Alonzo branch (3-element array, no `isValid`). Flag off: payloads stay CBOR maps as before.


stores/transaction/src/main/java/com/bloxbean/cardano/yaci/store/transaction/processor/TransactionProcessor.java
```diff
@@ -275,7 +275,12 @@ private void collectTransactionCbor(Transaction transaction, List<TxnCbor> txnCb
             return;
         }
 
-        String cborHex = transaction.getBody().getCbor();
+        //Full tx cbor is spliced from the original block bytes by Yaci. It's null when it can't be
+        //derived byte-exact, in which case fall back to body-only cbor.
```
satran004: @edridudi @adamcazes  @huylevt 

Need your opinion here. If the transaction-level CBOR from Yaci is `null` for some reason, should we fall back to the transaction body CBOR, or just set it to `null`?

Falling back to the transaction body CBOR may cause confusion, especially if the user has enabled the `setFullCbor` flag.

edridudi: If the user has enabled the setFullCbor flag,and for some reason yaci can't provide me with  the full cbor, I would prefer that it will throw an error rather than giving me the wrong cbor. We should be aware that something is wrong.

satran004: Thanks @edridudi. Agreed. For the API response, we can return an error or an empty response.

But in the transaction processor, I don’t think we should throw an error, as that would stop the sync process. Maybe we can log the error and continue the indexing process.

Alternatively, we already have an `error` table that tracks parsing errors. In this case, we could introduce a new error code such as DATA_MISSING_ERROR and add the tx_hash and reason to the error table.

Tracking an error in the error table is just a matter of publishing an event. https://github.com/bloxbean/yaci-store/blob/456b831094dd923ce075abbbe44d33ffa6262a09/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/BlockFetchService.java#L388

@Sotatek-HenryDo Can you please look into this?

adamcazes: Agreeing with @edridudi, and adding the measurements behind #1090 since they bear directly on the fallback question.

**Why a silent fallback is worse than an absent row.** The two payloads are both valid CBOR and are not distinguishable without decoding. On our mainnet instance, the same transaction is 453 bytes starting `0xa9` (a map, the body) against hosted Blockfrost's 1327 bytes starting `0x84` (array of 4: body, witness set, is_valid, auxiliary data). A consumer that deserializes the response as a transaction does not get "data missing", it gets a parse failure or, worse, a successful parse of something that is not the transaction it asked for.

**The `size` field is the part that has no tell.** `/txs/{hash}` `size` is read from `transaction_cbor.cbor_size`, so a body-only fallback row does not just affect `/txs/{hash}/cbor`, it silently makes `size` wrong on every affected transaction, and there is nothing in the response for a caller to notice. Measuring stored body-only sizes against real block body sizes on mainnet, the shortfall ran from 28 to 92 percent depending on witness weight. Blockfrost's `size` is the full transaction, so any fallback row puts us quietly out of parity on a field nobody will think to check.

**Suggestion, on top of the error table.** The error table gives operators visibility, which is right, but the Blockfrost API layer reads `transaction_cbor`, not the error log, so at read time it still cannot tell a full-cbor row from a fallback row. Two options that fix that:

1. Do not write a row when the full cbor is unavailable. `/txs/{hash}/cbor` then 404s and `size` is null, which is the behaviour consumers already handle today, and it is honest.
2. If a fallback row is useful for other consumers, add a discriminator on the row (for example a `cbor_kind` of `FULL` or `BODY`) so the API layer can serve full rows, and 404 or omit `size` for body-only ones.

We prefer option 1 for its simplicity. Either way, we agree the processor must not throw: never stop the sync for this, and `DATA_MISSING_ERROR` with the tx hash and reason is the right shape for the error table.

Happy to validate once this lands. We have the before state captured on a synced mainnet instance (the 453-byte body-only payload and Blockfrost's 1327-byte full transaction for the same tx), so the before and after comparison is ready to run, and we can report `size` parity against the live API at the same time.

Sotatek-HenryDo: Pushed a fix (bfb7b1348): when `saveFullTxCbor` is on and Yaci's `getTxCbor()` comes back null, we no longer fall back to body-only cbor. Instead we skip storing that row, log the gap, and publish a new `DATA_MISSING_ERROR` event (block + tx_hash) to the error table so the sync keeps going. Matches @adamcazes's option 1: `/txs/{hash}/cbor` 404s and size stays null for those transactions, no silent wrong-payload row.

Sotatek-HuyLe3a (approved):

Issue #896, opened 2026-04-09
Track Deposit Amount in `stake_registration`


Sotatek-HuyLe3a:
### Background

  Currently, `stake_registration` does not store a `deposit` field. The deposit amount is taken from protocol parameter.

  ### Problem

  Conway introduced new certificate types where a custom deposit coin can be specified explicitly by the submitter:

  - **`RegCert`** — Conway-era stake registration with explicit `coin` (deposit)
  - **`UnregCert`** — Conway-era stake deregistration with explicit `coin` (refund)
  - Related combined certs: `StakeRegDelegCert`, `VoteRegDelegCert`, `StakeVoteRegDelegCert`

PR #1111, opened 2026-08-24
feat(staking): persist stake-key deposit on stake_registration

### LINKED ISSUE #451: Track stake registration deposit in stake_registration table

huylevt:
## Summary

Stake-key deposit is no longer treated as a constant 2 ADA. Each `stake_registration` row now stores the amount paid or refunded at ingest. Conway certificates use their explicit `coin`; legacy registrations use the applicable epoch/genesis `keyDeposit`; legacy deregistrations recover the deposit recorded by the active registration instead of using the current protocol parameter.

Blockfrost `GET /accounts/{stake_address}/registrations` reads the stored value on registered rows and returns `null` on deregistered rows. Registered rows without a stored deposit use the latest available protocol `keyDeposit`, falling back to Shelley genesis when the epoch store is disabled.

### Schema and domain

- Add nullable `deposit bigint` through `V0_800_3__stake_registration_deposit.sql` for H2, PostgreSQL, and MySQL, keeping the original init migration checksum unchanged.
- Backfill existing `stake_registration` rows with the historical 2 ADA value used by Yaci Store before per-certificate deposits were persisted.
- Add `BigInteger deposit` to `StakeRegistrationEntity` and `StakeRegistrationDetail`.
- Include `sr.deposit` in `StakeRegistrationExporter`.

### Ingest

- Store the certificate `coin` for `REG_CERT`, `UNREG_CERT`, `STAKE_REG_DELEG_CERT`, `VOTE_REG_DELEG_CERT`, and `STAKE_VOTE_REG_DELEG_CERT`.
- Resolve legacy registration deposits from the epoch `keyDeposit`, then Shelley genesis, then the existing 2 ADA fallback.
- Resolve legacy deregistration refunds from the latest active registration lifecycle row ordered by slot, transaction index, and certificate index. Registrations already processed in the current event are tracked in memory because the batch has not yet been persisted. This matches `shelleyTotalRefundsTxCerts`, which uses a registration earlier in the transaction or looks up the credential's recorded deposit ([ledger source at revision `70bd9dbd`](https://github.com/IntersectMBO/cardano-ledger/blob/70bd9dbd6e365dc98c0a7c8981d3451ff2fbe2ba/eras/shelley/impl/src/Cardano/Ledger/Shelley/TxCert.hs#L632-L663)); `lookupDepositDState` reads that deposit from `depositAccountStateL` ([ledger source](https://github.com/IntersectMBO/cardano-ledger/blob/70bd9dbd6e365dc98c0a7c8981d3451ff2fbe2ba/libs/cardano-ledger-core/src/Cardano/Ledger/State/CertState.hs#L220-L224)).
- Keep stored certificate types as `STAKE_REGISTRATION` / `STAKE_DEREGISTRATION`.
- Use the same epoch/genesis fallback chain for pool deposits.
- Set `.deposit(getKeyDeposit(0))` on DevKit genesis synthetic rows in `GenesisPoolProcessor`.

### Genesis protocol-parameter fallback

- Use `ShelleyGenesisProtocolParamsProvider` in `components/common` to load and cache the complete Shelley genesis `ProtocolParams` rather than exposing only `keyDeposit`.
- Register the provider with `StakingStoreConfiguration` and reuse it from `DepositParamService` and the Blockfrost account service.
- Keep genesis values separate from `EpochParamClient` so static fallback data is not presented as the latest protocol parameters.
- Prefer a real `EpochParamClient` when one is available and use Shelley genesis only when the epoch client is absent.

### Blockfrost account API

- Select `stake_registration.deposit` in `findRegistrations`.
- Return the stored deposit string for registered rows, `null` for deregistered rows, and the latest epoch/genesis fallback for registered rows whose stored deposit is null.

### AdaPot deposit snapshots

- Calculate stake-key deposits from each `stake_registration.deposit` value instead of binding a constant 2 ADA protocol parameter.
- Add registration deposits and subtract deregistration refunds while continuing to exclude synthetic `Genesis` rows.
- Keep pool registration and retirement deposit accounting unchanged.

## Related Issues

- Closes #451
- Related to #1084 / #1101 (API workaround this PR replaces when the column is populated)
- Related to #896 (Conway certificate type names on the row — not in this PR)

## Testing

- [x] `./gradlew :components:common:test`
- [x] `./gradlew :stores:staking:test`
- [x] `./gradlew :aggregates:adapot:test`
- [x] `./gradlew :extensions:blockfrost:blockfrost-account:test`
- [x] `./gradlew :starters:spring-boot-starter:compileJava`
- [x] AdaPot snapshot test covers different registration and deregistration deposit values and synthetic Genesis exclusion.
- [x] Manual: Yaci DevKit with `keyDeposit=5000000` — genesis row and live registration both stored `5000000`; Blockfrost registrations returned `"5000000"`.

## Migration notes

- Existing installations receive the column and a 2 ADA historical backfill automatically through `V0_800_3`.
- Freshly ingested registrations and deregistrations store the deposit or refund resolved from the certificate and applicable ledger state.
- The backfill preserves Yaci Store's previous accounting behavior and matches public-network history covered by this change. A custom network whose historical stake-key deposit differed from 2 ADA requires a re-sync to reconstruct exact historical values.
- The column remains nullable so missing future ingest data is not silently converted to 2 ADA; AdaPot relies on the migration and normal ingest path to populate it.
- No new runtime configuration keys.

## Rollback safety

- Table `stake_registration` is already listed in `rollback.yml`.
- Rollback still deletes by slot; the new nullable column does not change rollback behavior.

## Out of scope

- Blockfrost transaction-level deposit aggregation


components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/genesis/ShelleyGenesisProtocolParamsProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.bloxbean.cardano.yaci.store.common.config.StoreProperties;
+import com.bloxbean.cardano.yaci.store.common.domain.ProtocolParams;
+import com.bloxbean.cardano.yaci.store.common.genesis.util.GenesisFileUtil;
+import com.bloxbean.cardano.yaci.store.common.util.StringUtil;
+
+import java.io.File;
+import java.util.Optional;
+
+public class ShelleyGenesisProtocolParamsProvider {
```
satran004: @huylevt Can we expose `ShelleyGenesis` object from `GenesisConfig` instead of creating a new helper ?

extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
```diff
@@ -47,26 +50,39 @@ public List<BFAccountDelegationDto> findDelegations(String stakeAddress, int pag
...
                 .stream()
                 .map(registration -> {
                     var dto = mapper.toRegistrationDto(registration);
-                    if (BFAccountMapper.ACTION_REGISTERED.equals(dto.getAction()))
-                        dto.setDeposit(keyDeposit);
+                    if (BFAccountMapper.ACTION_DEREGISTERED.equals(dto.getAction())) {
+                        dto.setDeposit(null);
+                    } else if (registration.deposit() != null) {
+                        dto.setDeposit(registration.deposit().toString());
+                    } else {
```
satran004, line 63: @huylevt This shouldn't be ideally null. right?

extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
```diff
@@ -47,26 +50,39 @@ public List<BFAccountDelegationDto> findDelegations(String stakeAddress, int pag
...
+        }
+
+        var genesisProvider = shelleyGenesisProtocolParamsProvider.getIfAvailable();
+        if (genesisProvider == null)
             return null;
 
-        return client.getLatestProtocolParams()
-                .map(ProtocolParams::getKeyDeposit)
+        return genesisProvider.getProtocolParams()
+                .map(com.bloxbean.cardano.yaci.store.common.domain.ProtocolParams::getKeyDeposit)
```
satran004: Let's use `ProtocolParams` instead of fqclass name if possible.

stores/staking/src/main/java/com/bloxbean/cardano/yaci/store/staking/service/DepositParamService.java
```diff
@@ -1,28 +1,33 @@
...
     }
 
     public BigInteger getPoolDeposit(int epoch) {
-        return protocolParamService.getProtocolParam(epoch)
-                .map(p -> p.getPoolDeposit())
-                .orElse(adaToLovelace(500));
+        return resolveDeposit(epoch, ProtocolParams::getPoolDeposit, adaToLovelace(500));
     }
 
     public BigInteger getDRepDeposit(int epoch) {
```
satran004: @huylevt Can we remove or fix the implementations of both getDRepDeposit and getGovActionDeposit? Currently, neither method is used.

stores/staking/src/main/java/com/bloxbean/cardano/yaci/store/staking/service/DepositParamService.java
```diff
@@ -32,4 +37,19 @@ public BigInteger getDRepDeposit(int epoch) {
...
+                    .map(getter)
+                    .orElse(null);
+            if (fromEpoch != null) {
+                return fromEpoch;
+            }
+        }
+
+        return shelleyGenesisProtocolParamsProvider.getProtocolParams()
+                .map(getter)
+                .orElse(hardcodedFallback);
```
satran004: Not sure if we should continue returning the hardcoded fallback here. Maybe we can keep it for now, but add a comment to remove it later.

extensions/blockfrost/account/src/main/java/com/bloxbean/cardano/yaci/store/blockfrost/account/service/BFAccountService.java
huylevt: I checked the Blockfrost OpenAPI contract. For account registration history, `deposit` is nullable and is explicitly `null` on `deregistered`  rows. We still persist the deregistration deposit internally, but the Blockfrost response should normalize it to `null`. I therefore kept this branch unchanged.

stores/staking/src/main/java/com/bloxbean/cardano/yaci/store/staking/service/DepositParamService.java
huylevt: Agreed. I confirmed that neither method has any caller in the repository. Since both implementations ignored the `epoch` argument and returned a constant, I removed them instead of preserving an unused and potentially misleading API

stores/staking/src/main/java/com/bloxbean/cardano/yaci/store/staking/service/DepositParamService.java
huylevt: I kept the fallback for now to avoid changing behavior for deployments where epoch parameters are unavailable. Epoch parameters are preferred first, followed by Shelley genesis parameters where applicable, so the constant is now only the final compatibility fallback. I renamed it to `legacyFallback` and documented the reason it remains. We can remove it once protocol-parameter availability is guaranteed for every supported configuration and define the expected failure behavior for missing parameters.

components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/genesis/ShelleyGenesisProtocolParamsProvider.java
huylevt: Good point about reusing the Shelley genesis already parsed by GenesisConfig. I avoided exposing the complete ShelleyGenesis object because it is currently mutable and also retains initialFunds and genesis staking  
 data, which GenesisConfig intentionally loads on demand. Instead, GenesisConfig now caches and exposes only the Shelley ProtocolParams required by these consumers. Exposing the full object would be reasonable later 
 if the genesis model is made immutable and multiple consumers genuinely need the complete document.

huylevt:
## Updates since `f42dcee`

  ### Why `f42dcee` was needed

  When block processing runs in parallel, block partitions may commit out of lifecycle order. A partition containing a legacy stake deregistration can therefore be persisted before another partition containing its
  preceding active registration.

  In that situation, the initial lookup cannot find the registration deposit, even though the registration exists in the valid chain history. This leaves the deregistration row with a `NULL` deposit.

  Commit `f42dcee` adds a post-batch reconciliation step:

  - After all partitions in a processed batch have completed, find legacy deregistrations in that batch whose deposit is still unresolved.
  - Resolve each deposit from the now-complete stake registration lifecycle history.
  - Update the existing deregistration rows rather than creating separate reconciliation state.
  - Keep the reconciliation compatible with rollback because the corrected deposit remains part of the original stake-registration row.
  - Fail if a deposit is still unresolved, because downstream deposit accounting requires complete deposit values.

  `BatchBlocksProcessedEvent` is used because reconciliation must happen only after all parallel partitions have completed. Running it from an individual certificate, pre-commit, or commit event would not guarantee that
  the registration from another partition is already visible.

  ### Review follow-up

  Commit `eb73e4889` addresses the review feedback:

  - Reuses Shelley protocol parameters already parsed by `GenesisConfig`.
  - Removes the duplicate `ShelleyGenesisProtocolParamsProvider` and its bean registration.
  - Exposes only cached Shelley `ProtocolParams` instead of retaining and exposing the complete `ShelleyGenesis` object.
  - Treats `GenesisConfig` as a required core dependency in both `DepositParamService` and `BFAccountService`, so missing core wiring fails fast.
  - Keeps `ProtocolParamService` and `EpochParamClient` optional because the epoch store may be disabled.
  - Uses Shelley genesis key deposit as the fallback when epoch protocol parameters are unavailable.
  - Removes the fully-qualified `ProtocolParams` reference from `BFAccountService`.
  - Preserves `deposit = null` for Blockfrost `deregistered` responses, as required by the Blockfrost OpenAPI contract, while retaining the persisted deposit internally for accounting.
  - Removes the unused `getDRepDeposit` and `getGovActionDeposit` methods after confirming that they have no callers in the repository.
  - Retains the legacy key/pool deposit constants only as the final compatibility fallback when neither epoch nor genesis parameters are available.
  - Adds and updates focused tests for `GenesisConfig`, deposit resolution, and Blockfrost account registration responses.

  ## Open question: fail-fast or continue syncing?

  The current reconciliation logic throws an exception when a legacy deregistration deposit remains unresolved after the entire batch has completed.

  This fail-fast behavior protects accounting correctness: continuing would leave a `NULL` deposit in `stake_registration`, and SQL aggregation may silently ignore that value, producing an incorrect deposit snapshot.

  Should an unresolved deregistration deposit remain a fatal invariant violation, or should the indexer:

  1. Publish an `ErrorEvent` containing the affected slot, transaction hash, and certificate index; and
  2. Continue syncing subsequent blocks?

  Continuing sync would improve availability, but we would also need to decide how to prevent `DepositSnapshotService` and other consumers from using incomplete deposit data. Possible safeguards include skipping
  affected snapshots, marking the store as degraded, or scheduling a durable reconciliation retry.

huylevt:
## Update since [dfdcf22](https://github.com/bloxbean/yaci-store/pull/1111/commits/dfdcf221eeb83ad93c0d2dc8c31fa84d4b8ae0c8)

## Stake Deregistration Deposit Handling

  ### Summary

  Stake certificate deposits are now stored according to the certificate data:

  - Legacy `STAKE_DEREGISTRATION` certificates are stored with a `NULL` deposit because they do not contain a coin value.
  - Conway `UNREG_CERT` certificates retain their explicit deposit.
  - Registration certificates use their explicit coin or the applicable key-deposit protocol parameter.

  ### Snapshot Resolution

  `DepositSnapshotService` resolves a legacy deregistration refund from the immediately preceding lifecycle row, ordered by:

  1. Slot
  2. Transaction index
  3. Certificate index

  The preceding lifecycle row must be a `STAKE_REGISTRATION`. This prevents the query from crossing an intervening deregistration and refunding the same registration twice.

  If the lifecycle deposit still cannot be resolved:

  - A `DATA_MISSING_ERROR` event is published and persisted for traceability.
  - The applicable epoch key-deposit protocol parameter is used as a defensive fallback.
  - AdaPot snapshot processing continues instead of failing.

  The fallback may differ from the original registration deposit on networks where the key-deposit parameter changed, but the persisted error makes the degraded calculation observable.

  ### Sync Path Simplification

  The sync-time deregistration reconciliation mechanism has been removed, including:

  - Per-event active-deposit caches.
  - `BatchBlocksProcessedEvent` reconciliation.
  - Per-certificate database lookups.
  - `getRegistrationBefore`.
  - `findUnresolvedDeregistrations`.

  This keeps stake certificate persistence aligned with the on-chain certificate data and moves lifecycle resolution to the epoch snapshot path.

  ### Deposit Parameter Resolution

  `DepositParamService` resolves both key and pool deposits using the following precedence:

  1. Protocol parameters for the requested epoch.
  2. Shelley genesis protocol parameters.
  3. Legacy defaults:
     - Key deposit: 2 ADA.
     - Pool deposit: 500 ADA.

  Regression coverage now verifies all three resolution paths for both `getKeyDeposit` and `getPoolDeposit`.

  ### Migration

  The stake registration deposit migration was updated consistently for H2, MySQL, and PostgreSQL:

  - Historical `STAKE_REGISTRATION` rows are backfilled with `2,000,000` lovelace.
  - Historical `STAKE_DEREGISTRATION` rows remain `NULL`.
  - Deregistration refunds are subsequently resolved by the snapshot lifecycle query.

  Networks with a different historical key deposit still require a resync for exact registration values.

  ### Verification

  - `StakeRegProcessorTest`: passed.
  - `DepositSnapshotServiceTest`: passed.
  - `DepositParamServiceTest`: passed.
  - Full `:aggregates:adapot:test`: passed.
  - The updated H2 migration was exercised successfully by the staking test task.
  - Added coverage for:
    - Legacy lifecycle resolution.
    - Certificate ordering within the same transaction.
    - Explicit Conway deregistration deposits.
    - Intervening deregistration lifecycles.
    - Multiple unresolved rows.
    - Protocol-parameter fallback and `DATA_MISSING_ERROR` publication.
    - Key and pool deposit resolution from epoch params, Shelley genesis, and legacy defaults.

satran004 (approved):

PR #1115, opened 2026-08-27
fix(docs): generate sitemap from docs pages and fix doubled slash in URLs

### LINKED ISSUE #1051: Sitemap URLs are doubled-slashed and point at removed routes
`store.yaci.xyz/sitemap.xml` lists 34 URLs and none of them resolve, so no documentation page is discoverable through it. Two separate things are going on.

**Doubled slash.** `DOCUMENTATION_URL` in `docs/utils/constants.ts` ends in a slash, and both `docs/app/sitemap.ts` and `docs/app/robots.ts` concatenate paths that begin with one. 33 of the 34 entries come out as `https://store.yaci.xyz//...`, and `robots.txt` advertises `//sitemap.xml`.

**Stale routes.** Fixing the slash alone does not help, because the paths are pre-versioning names. Not one sitemap entry contains `/docs/`, which is where the site serves everything now.

```
https://store.yaci.xyz//yacistore/design            404
https://store.yaci.xyz/yacistore/design             404
https://store.yaci.xyz/docs/v2/introduction/design  200
```

The same holds for the other 33.

The concatenation is a quick fix. What the routes should point at is the bigger question, and generating the list from `docs/app/docs/v*/` rather than maintaining it by hand may be worth considering, since a hand-written list is what let the two drift apart. Can open a PR for the slash part if that is useful.

Sotatek-HenryDo:
Fixes #1051.

`store.yaci.xyz/sitemap.xml` listed 34 URLs and none of them resolved. Two causes.

**Doubled slash.** `DOCUMENTATION_URL` ends in a slash and both `sitemap.ts` and `robots.ts` concatenated paths that begin with one, producing `https://store.yaci.xyz//...` and a `//sitemap.xml` entry in `robots.txt`. Both now resolve the path with `new URL(path, DOCUMENTATION_URL)`, which normalizes the join.

**Stale routes.** The hand-written list used pre-versioning paths (`/yacistore/design`, `/gettingstarted/...`), none of which contain `/docs/`, where the site has served everything since the v1/v2 split. Even with the slash corrected, all 34 were 404s, and three carried a `/yaciStore/` casing bug that a case-sensitive static host would not forgive.

The list is now generated from `app/docs/v2/**/page.mdx` rather than maintained by hand, since a hand-written list drifting from the docs tree is what caused this. `readdirSync(..., { recursive: true })` is used rather than `fs.globSync`, which needs Node 22 while CI runs Node 20.

v1 is deliberately excluded: it stays online, but v2 is the current version and listing both duplicates every page.

### Verification

Built with `yarn install --frozen-lockfile && yarn export`, matching CI, and checked the generated `out/`:

- 63 URLs, up from 34
- 0 URLs containing a doubled slash, down from 33
- 0 URLs without a corresponding exported HTML file, i.e. no 404s
- all 62 v2 pages present; the only non-page entry is the site root


Sotatek-HuyLe3a (approved):

Sotatek-HuyLe3a:
LGTM @Sotatek-HenryDo !

satran004 (approved):
LGTM. Thanks @Sotatek-HenryDo

Issue #1132, opened 2026-09-04
Add rollback operations for asset-ext store in rollback cli


Sotatek-HuyLe3a:
Need to check if new assets-ext table are in rollback command in admin-cli (at least cip68 table which reads from onchain data)

PR #1133, opened 2026-09-04
fix(admin-cli): add cip68 rollback table

huylevt:
## Summary

  - Add `cip68_metadata` to the default admin-cli rollback configuration.
  - Verify rollback config loading includes CIP-68 on-chain metadata rollback.
  - Verify rollback SQL generation deletes CIP-68 rows by `slot > :slot`.

  ## Details

  The asset-ext CIP-68 metadata table stores on-chain reference NFT metadata and already uses slot-based rollback in the runtime rollback processor. The admin-cli rollback path loads table operations from
  `rollback.yml`, so `cip68_metadata` must be listed there for manual rollback operations to clean rows after the rollback point.

  CIP-113 rollback is intentionally not added because the CIP-113 implementation is expected to be removed from asset-ext.

#1132

### DIFF
--- a/components/dbutils/src/main/resources/rollback.yml
+++ b/components/dbutils/src/main/resources/rollback.yml
@@ -6,6 +6,13 @@ rollback:
         type: "slot"
         column: "slot"
         operator: ">"
+
+    - name: "cip68_metadata"
+      operation: "DELETE"
+      condition:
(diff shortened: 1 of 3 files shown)


satran004 (approved):
LGTM

PR #1134, opened 2026-09-07
chore(assets-ext): remove CIP-113 programmable token support

Sotatek-HenryDo:
Removes the CIP-113 programmable token vertical from the Asset Store Extension.

Per the standup on 2026-09-04: CIP-113 upstream has moved to upgradable contracts, which
may change the on-chain design fundamentally. Rather than carry tables that would need a
migration later, the code is removed now, while the extension has not shipped yet.

The implementation is preserved on `feature/cip113-assets-ext` (branched from `main` at
`4a71c9c0f`) so it can be brought back once CIP-113 stabilises.

## No migration needed

`extensions/assets-ext` has never been part of a release tag — its first commit `fa9ddf267`
is not an ancestor of `v3.0.0-beta1/2/3`. `V0_1700_1__init.sql` is therefore unreleased and
is edited in place: no new Flyway version, no down-migration, nothing for operators to run.

## What is removed

- `cip113/` package (configuration, parser, processor, rollback processor, service, storage,
  entities, repository) and its tests
- `cip113_registry_node` table and `idx_cip113_slot` from the `postgresql`, `mysql` and `h2`
  copies of `V0_1700_1__init.sql`
- Per-network registry NFT policy ID files under `src/main/resources/cip113/`
- `Cip113CredentialType` from the GraalVM `reflect-config.json`
- `AssetsReader.getCip113RegistryNode` / `getCip113RegistryNodes` / `isProgrammableToken`
- The CIP-113 reader, batch prefetch map and extension assembly in `TokenQueryService`
  (batch prefetch drops from 4 queries to 3)
- `store.assets.ext.cip113.*` configuration in both the extension and the starter
- CIP-113 sections in the extension README, `db/store/README.md`, `config/application.properties`,
  `docker/config/application.properties` and `application-all.properties`

## API change

`Subject.type` and `Subject.extensions` are removed along with the `TokenType` enum and the
`Extension` interface. CIP-113 was the only producer of either — `type` was permanently
`NATIVE` and `extensions` permanently empty once CIP-113 was gone.

Both fields were yaci-store additions on top of the cf-token-metadata-registry response
shape, and `scripts/verify_against_cf_registry.py` compares as a subset, so CF parity is
unaffected.

## Optional indexes moved to `asset-ext-index.yml`

Review follow-up. The assets-ext optional indexes were the only ones in the repo shipped as
hand-written SQL for the operator to run; everywhere else they are declared as YAML under
`components/dbutils/src/main/resources/` and applied by the admin CLI.

`extensions/assets-ext/src/main/resources/db/store/optional-indexes.sql` is therefore deleted
and replaced by `components/dbutils/src/main/resources/asset-ext-index.yml`, following the
`blockfrost-index.yml` pattern — blockfrost has no `.sql` counterpart either:

```
apply-optional-indexes --indexes asset-ext-index
```

The name was already advertised in `DBCommands.applyOptionalIndexes` help text
(`--indexes 'blockfrost-index,asset-ext-index'`) but the file never existed, so this also
closes a dangling reference.

Both live indexes carry over — `idx_cip26_metadata_ticker` and
`idx_cip68_metadata_policy_label` (whose `slot DESC` column passes through `IndexService`
verbatim). The three commented-out PostgreSQL-only shapes in the old file (`pg_trgm` on
`cip26_metadata.name`, JSONB GIN on either `properties` column) cannot be expressed in the
index definition format — no operator-class field — so they survive as documented comments
in the YAML rather than being dropped silently.

`IndexLoaderTest.loadAssetExtIndexes` covers the new file, matching the existing per-file
tests. The YAML is otherwise never loaded during the build, so a syntax error in it would
not surface until an operator ran the CLI.

## Verification

- `./gradlew :components:dbutils:test :extensions:assets-ext:test :starters:assets-ext-spring-boot-starter:test` — pass
- `./gradlew classes testClasses` — all modules compile
- `git grep` for `cip113` / `CIP-113` / `programmable` across tracked files — no hits

Rebased onto `main` to pick up `ee0212cbc`, which added the assets extension section to the
docker config after this branch was cut; the CIP-113 lines it introduced are removed here so
the docker block matches `config/application.properties`.


extensions/assets-ext/src/main/resources/db/store/optional-indexes.sql
```diff
@@ -19,6 +19,3 @@ CREATE INDEX IF NOT EXISTS idx_cip68_metadata_policy_label ON cip68_metadata(pol
 
 -- CIP-68 NFT image / collection-attribute lookups via the JSONB column (PostgreSQL only):
 -- CREATE INDEX IF NOT EXISTS idx_cip68_metadata_properties ON cip68_metadata USING GIN(properties);
-
```
satran004, line 22: @Sotatek-HenryDo @huylevt 
Shouldn't these index go into an asset-ext specific optional index yml?

config/application.properties
```diff
@@ -346,19 +346,16 @@ logging.level.com.bloxbean.cardano.yaci.store.core.service=INFO
...
-# CIP-113 registry NFT policy IDs auto-detected from protocol-magic.
 #######################################################################################
 #store.assets.ext.enabled=true
 # CIP-26 and CIP-68 are enabled by default when the master flag is on. To disable one:
 #store.assets.ext.cip26.enabled=false
 #store.assets.ext.cip68.enabled=false
-# CIP-113 is off by default (not yet live on mainnet). To enable:
-#store.assets.ext.cip113.enabled=true
 
 spring.threads.virtual.enabled=true
```
satran004, line 361: @Sotatek-HenryDo We also need to do similar change in docker/config/application.properties

extensions/assets-ext/src/main/resources/db/store/optional-indexes.sql
Sotatek-HuyLe3a: I think we can add `asset-ext-index.yml` (same format as `blockfrost-index.yml`) for `idx_cip26_metadata_ticker` and `idx_cip68_metadata_policy_label` in this PR.

config/application.properties
Sotatek-HuyLe3a: Also,  add the CIP-26 + CIP-68 assets-ext block there so it matches `config/application.properties

matiwinnetou (approved):
LGTM

extensions/assets-ext/src/main/resources/db/store/optional-indexes.sql
Sotatek-HenryDo: Done. Added `asset-ext-index.yml` with `idx_cip26_metadata_ticker` and `idx_cip68_metadata_policy_label`, same pattern as `blockfrost-index.yml`.

config/application.properties
Sotatek-HenryDo: Done. ee0212cbc had already added the section to the docker config with the CIP-113 lines, so I rebased and removed them. `docker/config/application.properties` now matches `config/application.properties`.

Sotatek-HuyLe3a (approved):

PR #1135, opened 2026-09-07
fix(common): use byron genesis slotDuration for byron slot length

### LINKED ISSUE #1053: Byron-era block timestamps use a fixed 20s slot duration, ignoring byron-genesis blockVersionData.slotDuration
### Summary

When indexing a Cardano-protocol chain whose **Byron-era `slotDuration` is not 20s**, yaci-store computes block timestamps as if Byron slots were 20s, ignoring the `blockVersionData.slotDuration` in the supplied `byron-genesis.json`. Every post-Byron block timestamp is then off by a constant `byronSlots × (20 − actualByronSlotSeconds)`.

For standard Cardano (and chains that really do use 20s Byron slots) this is correct and unnoticed. It breaks for custom chains that use a shorter Byron slot.

### Environment

- yaci-store `2.0.2.1` (also reproduced on `2.0.0`), yaci-core `0.4.0`
- Reading from a local `cardano-node` via N2N, genesis supplied via `store.cardano.byron-genesis-file` / `shelley-genesis-file`

### Evidence (two chains, same yaci-store, differ only in Byron slotDuration)

| chain | `byron-genesis` `blockVersionData.slotDuration` | yaci tip `time` vs wall-clock |
|-------|--------------------------------------------------|-------------------------------|
| A | `"20000"` (20s) | correct (~0s) |
| B | `"1000"` (1s)  | **+820800s (~9.5 days in the future)** |

Chain B has 2 Byron epochs = 43200 Byron slots. yaci's startup log prints:

```
###### Genesis Config #####
ByronEra Slot Duration  : 20.0
```

`820800 = 43200 × (20 − 1)` — i.e. yaci timed B's 43200 Byron slots at 20s each instead of the genesis-declared 1s. The overshoot is a fixed constant for all subsequent blocks (the Byron era is closed).

The corresponding `cardano-node` reads the same genesis and reports the tip at real wall-clock time (`syncProgress 100%`), so the chain itself is fine — only yaci-store's slot→time conversion diverges.

### Expected

Byron-era slot duration should come from `byron-genesis.json` `blockVersionData.slotDuration` (milliseconds), not a fixed 20s, so block timestamps are correct for chains with non-standard Byron slot timing.

### Actual

Byron slot duration is effectively fixed at 20s (`ByronEra Slot Duration : 20.0` regardless of genesis), so timestamps on chains with a shorter Byron slot land in the future by `byronSlots × (20 − actualSeconds)`.

### Repro

1. Take any chain whose `byron-genesis` sets `blockVersionData.slotDuration` to something other than `20000` (e.g. `1000`) and that has ≥1 Byron epoch before the Shelley hard fork.
2. Sync yaci-store against it with that genesis.
3. Compare `GET /api/v1/blocks/latest` `time` to the node's tip time / wall-clock → constant future offset `= byronSlotCount × (20 − slotDuration/1000)` seconds.

Happy to provide the two genesis files / a minimal repro if useful.

satran004:
Fixes #1053

## Problem

`ByronGenesis.readGenesisData()` already converts `blockVersionData.slotDuration` from milliseconds to seconds, but `getByronSlotLength()` divided by 1000 a **second** time:

```java
byronSlotLength = ...get(ATTR_SLOT_DURATION).asLong() / 1000; //in second   -> 20

public long getByronSlotLength() {
    if (byronSlotLength == 0)
        return 0;
    else
        return byronSlotLength / 1000;   // 20 / 1000 == 0  (integer division)
}
```

The getter returns a non-zero value only when the field exceeds 1000, i.e. when the genesis `slotDuration` is at least 1,000,000 ms. No real chain declares that, so in practice the method always returned `0`, and `GenesisConfig.slotDuration(Era.Byron)` fell through to its hardcoded 20s fallback. That is why the startup banner printed `ByronEra Slot Duration : 20.0` even though the genesis file had been parsed correctly.

Credit to @edridudi for the root-cause analysis in the issue.

## Impact

Mainnet, preprod and preview all declare `slotDuration: 20000`, whose Byron slots genuinely are 20s — they computed `20000 → 20 → 0 → 20` and landed on the right answer by accident. **No behaviour change for those networks.**

A chain whose Byron era did not run at 20s had its Byron block times computed at 20s/slot. The error accrued at `(20 - actualSeconds)` per Byron slot and then froze once Byron closed, so every post-Byron block inherited it as a constant offset of `byronSlotCount × (20 − actualSeconds)` seconds. In the reported case (`slotDuration: 1000`, 43200 Byron slots) that is `43200 × 19 = 820800`s — timestamps ~9.5 days in the future.

## Fix

Drop the explicit getter so Lombok's `@Data` getter returns the already-converted field. Same `long` signature, so this backports cleanly to `release-2.0.x` (the issue was reported against 2.0.2.1).

The 20s fallback in `GenesisConfig.slotDuration(Era.Byron)` is deliberately kept for genesis files that fail to supply the field — it just stops being reached for well-formed genesis.

## Note for affected chains

`block_time` is computed at index time and stored, so **this patch does not repair already-indexed data**. A chain with a non-20s Byron slot keeps the offset on every existing row; only newly indexed blocks come out right. Those deployments need a resync from genesis (or a rollback to before the Byron/Shelley boundary). Mainnet/preprod/preview deployments need nothing.

## Tests

- `ByronGenesisTest.getByronSlotLength_mainnet` → 20 and `getByronSlotLength_custom` → 1, the latter using the already-present-but-unused `custom-byron-genesis.json` (`slotDuration: 1000`) as the regression vector for this issue.
- `GenesisConfigTest.slotDuration_byron_ReturnsValueFromGenesis` guards the integration path — asserts the value now comes from the genesis rather than the fallback.

Both fail against the old code (`20 / 1000 == 0`).

## Other genesis subclasses

Checked, per the reporter's suggestion — no other instance of this shape. `ShelleyGenesis` reads `slotLength` with `asDouble()` and no division; Alonzo and Conway have no slot fields.

## Known limitation left as-is

`readGenesisData()` still truncates ms→s with integer division, so a sub-second Byron `slotDuration` would floor to `0` and hit the fallback. That predates this bug, no known chain uses one, and fixing it would mean changing the public getter's return type to `double` — which would block the 2.0.x backport.

### DIFF
--- a/components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/genesis/ByronGenesis.java
+++ b/components/common/src/main/java/com/bloxbean/cardano/yaci/store/common/genesis/ByronGenesis.java
@@ -36,7 +36,7 @@ public class ByronGenesis extends GenesisFile {
     private List<GenesisBalance> avvmGenesisBalances;
     private List<GenesisBalance> nonAvvmGenesisBalances;
     private long startTime;
-    private long byronSlotLength;
+    private long byronSlotLength; //in second
     private long protocolMagic;
     private long k;
 
@@ -50,13 +50,6 @@ public class ByronGenesis extends GenesisFile {
 
     public ByronGenesis(long protocolMagic) {
         super(protocolMagic);
-    }
-
-    public long getByronSlotLength() {
-        if (byronSlotLength == 0)
-            return 0;
-        else
-            return byronSlotLength / 1000;
     }
 
     public long getEpochLength() {
--- a/components/common/src/test/java/com/bloxbean/carano/yaci/store/common/genesis/ByronGenesisTest.java
+++ b/components/common/src/test/java/com/bloxbean/carano/yaci/store/common/genesis/ByronGenesisTest.java
@@ -23,6 +23,7 @@ class ByronGenesisTest {
 
     public static final String GENESIS_MAINNET_BYRON_GENESIS_JSON = "/genesis/mainnet-byron-genesis.json";
     public static final String GENESIS_PREPROD_BYRON_GENESIS_JSON = "/genesis/preprod-byron-genesis.json";
+    public static final String GENESIS_CUSTOM_BYRON_GENESIS_JSON = "/genesis/custom-byron-genesis.json";
 
     ObjectMapper objectMapper = new ObjectMapper();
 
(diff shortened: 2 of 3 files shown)


huylevt (approved):

huylevt:
LGTM @satran004 !

PR #1139, opened 2026-09-08
fix(governance): select latest spo and committee votes

### LINKED ISSUE #1069: [Governance] getVotesBySPO and getVotesByCommittee can return duplicate rows per voter+proposal (max(SLOT) join not migrated to ranked query)
## Summary

`VotingAggrService.getVotesBySPO` and `VotingAggrService.getVotesByCommittee` still use the
`max(SLOT)` + self-join shape that `e13fbc392` ("fix: ignore stale DRep votes after
unregistration") replaced on the DRep path. That shape does not guarantee **one row per
`(voter, gov_action)`**: when a voter has more than one vote row at the same maximum slot for
the same governance action, the join returns every one of them.

Only `getVotesByDRep` was migrated to the ranked form. The SPO and Committee paths were left
on the old query.

## Where

`aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java`

| Method | Line (on `main`) | Shape |
|---|---|---|
| `getVotesBySPO` | 51 | `max(SLOT)` self-join — join predicate at line 92 |
| `getVotesByCommittee` | 114 | `max(SLOT)` self-join — join predicate at line 161 |
| `getVotesByDRep` | 177 | ranked, `vote_rank = 1` at line 251 — **already fixed** |

The current SPO join key is:

```java
.on(VOTING_PROCEDURE.VOTER_HASH.eq(vVoterHash)
        .and(VOTING_PROCEDURE.GOV_ACTION_TX_HASH.eq(vGovActionTxHash))
        .and(VOTING_PROCEDURE.GOV_ACTION_INDEX.eq(vGovActionIndex))
        .and(VOTING_PROCEDURE.SLOT.eq(vMaxSlot)))
```

`slot` alone is not unique. The `voting_procedure` primary key is
`(tx_hash, voter_hash, voter_type, gov_action_tx_hash, gov_action_index)`, so two rows for the
same voter and the same governance action can share a slot whenever the votes land in two
different transactions inside the same block.

`e13fbc392` fixed exactly this for DReps by ranking on the full position tuple:

```java
rowNumber()
    .over(partitionBy(VOTER_HASH, VOTER_TYPE, GOV_ACTION_TX_HASH, GOV_ACTION_INDEX)
          .orderBy(SLOT.desc(), TX_INDEX.desc(), IDX.desc()))
    .as("vote_rank")
...
.and(vVoteRank.eq(1))
```

### SPO path

`SPOVotingDataCollector.collectSPOVotes` calls `calculateSPOStakeByVote` once per vote type
(`SPOVotingDataCollector.java:103-105`), and that method sums `epoch_stake` for every pool id
in the list without de-duplicating:

```java
var poolIds = votes.stream()
        .filter(vote -> vote.getVote().equals(voteType))
        .map(VotingProcedure::getVoterHash)
        .toList();
```

With two surviving rows for one pool:

- if both rows carry the same vote, the pool's stake is counted twice inside that bucket;
- if the two rows carry **different** votes (e.g. an earlier `NO` and a later `YES` in the same
  block), the pool's stake is counted in **both** the yes and no buckets.

Either way the SPO approval ratio is skewed. In the second case the over-counting also flows
into `totalDoNotVoteStake`:

```java
BigInteger totalDoNotVoteStake = totalStake
        .subtract(yesVoteStake)
        .subtract(noVoteStake)
        .subtract(abstainVoteStake)
        .subtract(delegateToAutoAbstainDRepStake)
        .subtract(delegateToNoConfidenceDRepStake);
```

which can go **negative** and be persisted into `gov_action_proposal_status.voting_stats`
(`VotingStatsService` writes `spoDoNotVoteStake` unclamped). Note this becomes more reachable
once #1034 lands, because the per-proposal default buckets it computes are systematically
larger than the previous batch-global ones, leaving less headroom in the subtraction.

Also worth deciding as part of this: the ledger's rule is "the last vote wins". Picking the
correct single row is the fix; the duplicate is a symptom.

### Committee path

`VotingDataCollector.collectCommitteeVotes` collapses the rows into a map with a
first-wins merge (`VotingDataCollector.java:149-154`):

```java
Map<String, Vote> votes = committeeVotes.stream()
    .collect(Collectors.toMap(
            VotingProcedure::getVoterHash,
            VotingProcedure::getVote,
        (existing, replacement) -> existing
    ));
```

The result set has no `ORDER BY`, so "existing" is whichever row the database happened to
return first. Two same-slot committee votes therefore resolve **non-deterministically**, and
the chosen vote may be the superseded one rather than the last one. This does not crash, but it
can silently flip a committee yes/no count and is not reproducible across runs.

## Suggested fix

Apply the same `rowNumber()` ranking used by `getVotesByDRep` to `getVotesBySPO` and
`getVotesByCommittee`:

- partition by `(voter_hash, voter_type, gov_action_tx_hash, gov_action_index)`
- order by `slot desc, tx_index desc, idx desc`
- keep `vote_rank = 1`

The existing extra predicates should be preserved as-is:

- SPO: the `EXISTS (epoch_stake …)` guard that restricts votes to pools with stake in the epoch;
- Committee: the hot-key filter.

Note the DRep query also gained an `unregistrationAtOrAfterVote` `NOT EXISTS` clause in
`e13fbc392`. That part is DRep-specific (a DRep unregistration clears its earlier votes) and
should **not** be copied to the SPO/Committee paths — only the ranking change applies here.

## Suggested tests

- `VotingAggrServiceTest`: two SPO votes for the same pool and proposal at the same slot with
  different `tx_index`, asserting only the higher-`tx_index` vote is returned.
- Same for two votes in the same transaction, differing only by `idx`.
- Committee equivalent, asserting the returned vote is the last one rather than an arbitrary
  one.
- A `SPOVotingDataCollector` case asserting `doNotVoteStake` stays non-negative when a pool
  has multiple same-slot vote rows.

## References

- `e13fbc392` — `fix: ignore stale DRep votes after unregistration` (introduced the ranked query
  on the DRep path only)
- Found while reviewing #1034, which changes SPO default-vote classification to be
  proposal-scoped and touches the same tally path.

huylevt:
## Summary

- Ensure SPO and Constitutional Committee vote aggregation returns only the latest vote for each voter and governance action.
- Resolve same-slot votes deterministically using transaction and vote ordering.
- Add regression coverage for both voter groups.

Fixes #1069

## Root cause

`VotingAggrService.getVotesBySPO` and `getVotesByCommittee` selected the maximum vote slot and joined it back to `voting_procedure`. A slot does not uniquely identify a vote, so multiple votes from the same voter for the same governance action could survive when they were submitted in the same block.

This could double-count SPO stake across voting buckets and made Committee vote selection depend on database result order.

## Changes

- Replace the `max(slot)` self-joins in the SPO and Committee queries with ranked vote CTEs.
- Partition votes by voter hash, voter type, governance action transaction hash, and governance action index.
- Rank votes by `slot DESC`, `tx_index DESC`, and `idx DESC`, and return only `vote_rank = 1`.
- Preserve the existing SPO epoch-stake eligibility check and Committee hot-key filter.
- Add tests covering `tx_index` and `idx` tie-breaking for SPO and Committee votes.

## Verification

- `JAVA_HOME=/home/huy/.jdks/openjdk-21 ./gradlew :aggregates:governance-aggr:test --tests com.bloxbean.cardano.yaci.store.governanceaggr.service.VotingAggrServiceTest`
- `JAVA_HOME=/home/huy/.jdks/openjdk-21 ./gradlew :aggregates:governance-aggr:test`
- `git diff --check`

All checks passed.

## Deployment validation

- [ ] Deploy and test the change on Preview. (in progress)
- [x] Deploy and test the change on Preprod.

## Notes

- No database schema changes are required.
- The DRep-specific unregistration filtering remains unchanged and is not applied to SPO or Committee votes.

### DIFF
--- a/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java
+++ b/aggregates/governance-aggr/src/main/java/com/bloxbean/cardano/yaci/store/governanceaggr/service/VotingAggrService.java
@@ -57,39 +57,49 @@ public class VotingAggrService {
                 .map(id -> row(id.getTransactionId(), id.getGov_action_index()))
                 .collect(Collectors.toList());
 
-        var cteQuery = select(
+        var rankedVoteQuery = select(
+                VOTING_PROCEDURE.TX_HASH.as("tx_hash"),
                 VOTING_PROCEDURE.VOTER_HASH.as("voter_hash"),
                 VOTING_PROCEDURE.VOTER_TYPE.as("voter_type"),
                 VOTING_PROCEDURE.GOV_ACTION_TX_HASH.as("gov_action_tx_hash"),
                 VOTING_PROCEDURE.GOV_ACTION_INDEX.as("gov_action_index"),
-                max(VOTING_PROCEDURE.SLOT).as("max_slot")
+                rowNumber()
+                        .over(partitionBy(
+                                VOTING_PROCEDURE.VOTER_HASH,
+                                VOTING_PROCEDURE.VOTER_TYPE,
+                                VOTING_PROCEDURE.GOV_ACTION_TX_HASH,
+                                VOTING_PROCEDURE.GOV_ACTION_INDEX
+                        ).orderBy(
+                                VOTING_PROCEDURE.SLOT.desc(),
+                                VOTING_PROCEDURE.TX_INDEX.desc(),
+                                VOTING_PROCEDURE.IDX.desc()
+                        ))
+                        .as("vote_rank")
         )
                 .from(VOTING_PROCEDURE)
                 .where(VOTING_PROCEDURE.EPOCH.le(epoch))
                 .and(row(VOTING_PROCEDURE.GOV_ACTION_TX_HASH, VOTING_PROCEDURE.GOV_ACTION_INDEX).in(govActionPairs))
-                .and(VOTING_PROCEDURE.VOTER_TYPE.eq("STAKING_POOL_KEY_HASH"))
-                .groupBy(
-                        VOTING_PROCEDURE.VOTER_HASH,
-                        VOTING_PROCEDURE.VOTER_TYPE,
-                        VOTING_PROCEDURE.GOV_ACTION_TX_HASH,
-                        VOTING_PROCEDURE.GOV_ACTION_INDEX
-                );
+                .and(VOTING_PROCEDURE.VOTER_TYPE.eq("STAKING_POOL_KEY_HASH"));
 
-        var voteWithMaxSlot = table(name("vote_with_max_slot"));
+        var rankedVote = table(name("ranked_vote"));
 
-        var vVoterHash = field(name("vote_with_max_slot", "voter_hash"), String.class);
-        var vGovActionTxHash = field(name("vote_with_max_slot", "gov_action_tx_hash"), String.class);
-        var vGovActionIndex = field(name("vote_with_max_slot", "gov_action_index"), Integer.class);
-        var vMaxSlot = field(name("vote_with_max_slot", "max_slot"), Long.class);
+        var vTxHash = field(name("ranked_vote", "tx_hash"), String.class);
+        var vVoterHash = field(name("ranked_vote", "voter_hash"), String.class);
+        var vVoterType = field(name("ranked_vote", "voter_type"), String.class);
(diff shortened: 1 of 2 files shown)


satran004 (approved):
LGTM. Thanks @huylevt

PR #1140, opened 2026-09-09
feat: add tx_index for exact intra-block ordering

### LINKED ISSUE #1033: Add tx_index (and cert_index where relevant) to tables lacking exact intra-block ordering
## Problem

Several tables only store `slot`, which isn't fine-grained enough to reproduce Blockfrost's exact ordering when multiple txs/certs share the same slot. Flagged in [PR #836 review](https://github.com/bloxbean/yaci-store/pull/836#discussion_r3056813952) and the Blockfrost gap docs in #958.

## Tables needing `tx_index` (and/or `cert_index`)

| Table | Currently has | Source |
|---|---|---|
| `withdrawal` | `tx_hash`, `slot` | [PR #836 comment](https://github.com/bloxbean/yaci-store/pull/836#discussion_r3056813952) |
| `address_utxo` | `slot`, `block` | [address_gaps.md#L63](https://github.com/bloxbean/yaci-store/blob/c05e6c4aaa1747f999a2c2d7214ebd42ae03bd59/extensions/blockfrost/adr/address_gaps.md#L63) |
| `transaction_metadata` | `slot` only | [metadata_gaps.md#L45](https://github.com/bloxbean/yaci-store/blob/c05e6c4aaa1747f999a2c2d7214ebd42ae03bd59/extensions/blockfrost/adr/metadata_gaps.md#L45) |

## Suggested next step

Add `tx_index` (and `cert_index` for cert-bearing tables) at write time.

Sotatek-HenryDo:
Closes #1033

## What

Adds `tx_index` to the three tables that only carried `slot`, so transactions sharing a block can be ordered the way Blockfrost orders them.

| Table | Module | Migration |
|---|---|---|
| `address_utxo` | `stores/utxo` | `V0_200_2` |
| `withdrawal` | `stores/transaction` | `V0_300_4` |
| `transaction_metadata` | `stores/metadata` | `V0_600_2` |

Each migration is an additive nullable `ADD COLUMN`, present in all three vendor folders.

## Convention

`tx_index` is the transaction's position in its block's transaction list, **counting invalid transactions**. This matches the existing `transaction.tx_index` and, importantly, matches db-sync's `tx.block_index` — verified read-only against mainnet db-sync on four blocks containing phase-2 failures, where the invalid tx occupies an index and the sequence stays contiguous. Since Blockfrost is built on db-sync, this keeps our ordering aligned with it.

## Notes per table

- **`address_utxo`** — the domain object and processors already set `txIndex`; only persistence was missing. Note the write path is jOOQ (`UtxoStorageImpl.saveUnspent`), not the JPA mapper, so the column had to be added to the insert and on-duplicate-update as well as the entity. `GenesisUtxoProcessor` leaves it null, which is correct: genesis UTXOs are not produced by a block transaction. `RocksDBUtxoStorage` serializes the domain object and needs no change.
- **`withdrawal`** — processor loop changed to an indexed loop. Plain JPA + MapStruct write path.
- **`transaction_metadata`** — needed event plumbing: `TxAuxData` gains `txIndex`, set at both `ShelleyBlockEventPublisher` publish sites. The existing streams filter out invalid and null-auxdata transactions *before* mapping, so the index is now taken from the full list first. `RemoteEventPublisher` forwards the list unchanged.

## Not in this PR

- **No backfill.** Existing rows keep `tx_index = NULL`. **:warning: Reminder: the release note needs a section telling operators that `tx_index` is populated from the deploy point onward, and that historical coverage requires a resync.**
- **Analytics export untouched** — every exporter and the `address_utxo_flattened` view use explicit column lists, so a new source column cannot change exported output.
- **`cert_index`** — out of scope; all three tables are transaction-level, and the cert-bearing tables already have it.
- **Reading `tx_index` for ordering** in the Blockfrost-compatible endpoints is a follow-up. That code must tolerate `NULL` on a non-resynced deployment rather than assume the column is populated.

## Testing

- `./gradlew build` green: 1264 tests, 0 failures.
- New tests pin the invalid-transaction rule: `WithdrawalProcessorTest.givenInvalidTransactionBeforeValidOne_shouldCountItInTxIndex` and `ShelleyBlockEventPublisherTest.shouldSetAuxDataTxIndexCountingInvalidTransactions`.
- Preprod, ~350 blocks: `address_utxo` 3740/3740, `withdrawal` 273/273, `transaction_metadata` 80/80 all match `transaction.tx_index`. Pre-change rows correctly stayed NULL (27 withdrawals, 16 metadata), and both migrations applied out-of-order to that already-populated DB.
- Spot-checked against Koios preprod: metadata row at `tx_index` 10 of an 11-tx block, withdrawal at `tx_index` 4 — both match the chain order.

## Known gaps

- `publishBlockEventsInParallel` has no unit test (the mocked executor never runs the task). It was exercised during the preprod catch-up, where `parallelMode` is active, but by evidence rather than by a test.
- Mainnet db-sync value comparison per the deploy checklist still needs the change deployed to mainnet.


Sotatek-HuyLe3a (approved):

stores/metadata/src/main/resources/db/store/h2/V0_600_2__add_tx_index_to_transaction_metadata.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE transaction_metadata ADD COLUMN tx_index INT;
```
Sotatek-HuyLe3a, line 1: As v3 requires a full resync and this write path always supplies tx_index, should this column be INT NOT NULL (like most of tx_index field of other tables)? This would enforce the invariant for withdrawal and transaction_metadata. address_utxo should remain  nullable for genesis UTxOs. @satran004 @Sotatek-HenryDo

satran004: @Sotatek-HuyLe3a As we are not planning to force a full resync for this release, and we are not using these `tx_index` fields in any of the BF queries yet, I think we can keep them nullable for this release.

For the next release, we can make them non-nullable, enforce a full resync, and possibly merge this migration script with the table creation script. Can we create an issue to track this?

satran004 (approved):
Thanks @Sotatek-HenryDo LGTM

